PR #38, opened 2015-01-27
StatusCodePages + MVC sample

Tratcher:
https://github.com/aspnet/Diagnostics/issues/66

### DIFF
--- a/samples/Diagnostics.StatusCodes.Mvc/Controllers/ErrorsController.cs
+++ b/samples/Diagnostics.StatusCodes.Mvc/Controllers/ErrorsController.cs
@@ -1,0 +1,15 @@
+﻿using Microsoft.AspNet.Mvc;
+
+namespace Diagnostics.StatusCodes.Mvc.Controllers
+{
+    public class ErrorsController : Controller
+    {
+        // This controller is called to generate response bodies for 400-599 status codes from
+        // other locations in the app.
+        [Route("errors/{id:int}")]
+        public IActionResult Index(int id)
+        {
+            return new ObjectResult("Error page, status code: " + id) { StatusCode = id };
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


samples/Diagnostics.StatusCodes.Mvc/Controllers/ErrorsController.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.AspNet.Mvc;
+
+namespace Diagnostics.StatusCodes.Mvc.Controllers
+{
+    public class ErrorsController : Controller
```
Eilon: Should we litter a few comments around here to say what this is? I'll proposed some comments:

`// This controller is called to handle application errors`

samples/Diagnostics.StatusCodes.Mvc/Controllers/HomeController.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.AspNet.Mvc;
+
+namespace Diagnostics.StatusCodes.Mvc.Controllers
+{
+    public class HomeController : Controller
+    {
+        // GET: /<controller>/
+        public string Index()
+        {
+            return "Hello World, try /bob";
```
Eilon: Does this work because `/bob` causes a 404 or 500 or something and thus an error is shown? Might want to clarify that.

samples/Diagnostics.StatusCodes.Mvc/Startup.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.AspNet.Routing;
+using Microsoft.Framework.DependencyInjection;
+
+namespace Diagnostics.StatusCodes.Mvc
+{
+    public class Startup
+    {
+        public void Configure(IApplicationBuilder app)
+        {
+            app.UseStatusCodePages(new StatusCodePagesOptions().WithReExecute("/errors/{0}"));
```
Eilon: Proposed comment:

`// Register a handler for handling status code based error page. This example ends up using MVC to show an error page.`

Eilon:
:shipit: on this too.

davidfowl:
@Tratcher what's the deal with this?

PR #54, opened 2015-04-08
React to aspnet/Mvc@f9d53e3

dougbu:
- build break


dougbu:
fyi only; will push immediately to resolve build break
/cc @pranavkm

PR #56, opened 2015-04-15
Fixing the entropy solution

Praburaj:
Looks like there is no xproj file found for Serilog sample though the sln file includes it.

@muratg @Tratcher


Tratcher:
:shipit:

PR #67, opened 2015-08-15
Add an ICompileModule sample

Tragetaschen:
As suggested in aspnet/dnx#891, here is a sample for using an ICompileModule.

I don't have VS available here at home (because I'm lazy :) ), so the *.xproj and an updated solution aren't yet available. I'll do that on Monday.

Just a friendly reminder, because it was really annoying to implement the code in this PR: aspnet/dnx#2294


davidfowl:
Thanks

PR #74, opened 2015-11-20
Remove dependency on MVC's invoker implementation

rynowak:
Choosing to reimplement some of this code rather than inherit. The invoker
is really an implementation of MVC, and this sample frequently breaks.


rynowak:
FYI only

PR #75, opened 2015-11-24
reacting to new IServer IHttpApplication design

JunTaoLuo:
Reacting to https://github.com/aspnet/Hosting/pull/463

### DIFF
--- a/samples/Owin.Nowin.HelloWorld/NowinServer.cs
+++ b/samples/Owin.Nowin.HelloWorld/NowinServer.cs
@@ -2,9 +2,7 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Hosting.Server;
-using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Features;
-using Microsoft.AspNet.Http.Internal;
 using Microsoft.AspNet.Owin;
 using Nowin;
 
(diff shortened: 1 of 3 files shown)


samples/Owin.Nowin.HelloWorld/NowinServer.cs
```diff
@@ -44,7 +55,7 @@ public void Dispose()
 
         private Task HandleRequest(IDictionary<string, object> env)
         {
-            return _callback(new DefaultHttpContext(new OwinFeatureCollection(env)));
+            return _callback(new OwinFeatureCollection(env));
```
JunTaoLuo: Note that this doesn't actually work since `OwinFeatureCollection` is readonly and `RequestServicesContainerMiddleware` throws when trying to add the `IServiceProvidersFeature`. I bypassed this by checking the `IsReadOnly` property on the `HttpContext.Feature` but we might want to see if there's something different we want to do. @Tratcher

JunTaoLuo:
@Tratcher @davidfowl @lodejard

samples/Owin.Nowin.WebSockets/NowinServer.cs
```diff
@@ -27,9 +25,22 @@ public NowinServer(ServerBuilder serverBuilder)
             _serverBuilder = serverBuilder;
         }
 
-        public void Start(RequestDelegate requestDelegate)
+        public void Start<THttpContext>(IHttpApplication<THttpContext> application)
         {
-            _callback = requestDelegate;
+            _callback = async features =>
```
Tratcher: Doesn't this have the same problem you had in WebListener? Does NOwin implement the OnStarting callback? (owin.OnSendingHeaders I think). That may not be called until after DisposeContext.

JunTaoLuo: Going to add a comment for this

Tratcher:
:shipit:

Issue #81, opened 2015-12-19
Owin.OwinWebSockets sample broken


JunTaoLuo:
An internal error occurs when running this sample. Using diagnostics I see:
```
System.Collections.Generic.KeyNotFoundException: websocket.Accept
   at Microsoft.AspNet.Owin.OwinEnvironment.System.Collections.Generic.IDictionary<System.String,System.Object>.get_Item(String key)
   at SelfHostServer.Startup.<>c__DisplayClass0_0.<<Configure>b__2>d.MoveNext() in C:\gh\Entropy\samples\Owin.OwinWebSockets\Startup.cs:line 47
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNet.Owin.WebSocketAcceptAdapter.<>c__DisplayClass6_0.<<AdaptWebSockets>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNet.Diagnostics.DeveloperExceptionPageMiddleware.<Invoke>d__7.MoveNext()
```

Seems like we need to handle the case when the key doesn't even exist.

muratg:
@JunTaoLuo is this on CoreCLR or CLR/Mono or both? 

@Tratcher any ideas?

JunTaoLuo:
I tested this on CLR previously, did not confirm on CoreCLR but I presume the same issue will occur.

JunTaoLuo:
The issue was that https://github.com/aspnet/Entropy/blob/dev/samples/Owin.OwinWebSockets/Startup.cs#L45 fails since the key `websocket.Accept` does not exist. Possibly, we just need to check for existence first using `env.Contains` or use `env.TryGetValue` instead of using the indexer.

Tratcher:
I think the dictionary use to return null for missing entries but now throws. Change to TryGetValue.

JunTaoLuo:
merged to dev

PR #84, opened 2016-01-08
Handle non-websocket requests without crashing in sample

JunTaoLuo:
Fixes https://github.com/aspnet/Entropy/issues/81. Tested locally to verify that the sample no longer crashes for non-websocket requests.


JunTaoLuo:
cc @Tratcher

Tratcher:
:shipit:

PR #85, opened 2016-01-13
Added sample to demonstrate rendering view to a string

ajaybhargavb:
@Eilon @rynowak @pranavkm


samples/Mvc.RenderViewToString/EmailGraphHelper.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Mvc.RenderViewToString
+{
+    public class EmailReportHelper
```
dougbu: @rynowak how many negative #AllRyans points is naming this a `BlahHelper` worth?

samples/Mvc.RenderViewToString/Program.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class Program
+    {
+        public static void Main()
+        {
+            // Initialize the necessary services
+            var services = new ServiceCollection();
+            ConfigureDefaultServices(services, path: null);
+
+            // Add a custom service that is used in the view.
+            services.AddSingleton<EmailReportHelper>();
```
dougbu: Wait, is it a helper or a service? Why not just `EmailReporter`?

samples/Mvc.RenderViewToString/Views/EmailPartial.cshtml
```diff
@@ -0,0 +1,4 @@
+﻿@inject Mvc.RenderViewToString.EmailReportHelper ReportHelper
+
+    This is the partial view content.<br />
+    @ReportHelper.GenerateReport(Model.UserData1, Model.UserData2).<br />
```
dougbu: Would a `ViewComponent` work here? If so, that would show a bit more working e2e.

samples/Mvc.RenderViewToString/RazorViewHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+            ICompositeViewEngine viewEngine,
+            ITempDataProvider tempDataProvider,
+            IServiceProvider serviceProvider)
+        {
+            _viewEngine = viewEngine;
+            _tempDataProvider = tempDataProvider;
+            _serviceProvider = serviceProvider;
+        }
+
+        private ActionContext GetActionContext()
```
dougbu: Please move `private` method after `public` ones.

samples/Mvc.RenderViewToString/Program.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            var provider = services.BuildServiceProvider();
+            var helper = provider.GetRequiredService<RazorViewHelper>();
```
dougbu: Very strange to add something to DI where the only reference is a few lines earlier. Cleaner to use `ActivatorUtlities.CreateInstance<T>(IServiceProvider, params object[])`.

samples/Mvc.RenderViewToString/RazorViewHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public string RenderViewToString<TModel>(string name, TModel model)
+        {
+            var actionContext = GetActionContext();
+
+            var viewEngineResult = _viewEngine.FindView(actionContext, name, false);
+
+            if (!viewEngineResult.Success)
+            {
+                return "View not found";
```
dougbu: This looks too much like success to the caller. Either `bool TryRenderViewToString(string, TModel, out string)` or `throw` here.

dougbu:
:watch: (like the overall approach)

samples/Mvc.RenderViewToString/Program.cs
ajaybhargavb: Just to demonstrate `@inject` works.

dougbu: If that's interesting, suggest using a more mainline scenario for `@inject` then e.g. a `ViewLocalizer`. But that might add to many unrelated complications to the sample. And MVC uses `@inject` internally so you're not really testing anything new except this helper itself.

samples/Mvc.RenderViewToString/EmailGraphHelper.cs
rynowak: All the points I had out are always negative.

rynowak: Meta-question: what is this class intending to demonstrate? What's it a surrogate for in a 'real' application?

samples/Mvc.RenderViewToString/Program.cs
rynowak: I'd make the opposite suggestion. Let's not show `ActivatorUtilities` in a sample.

samples/Mvc.RenderViewToString/Program.cs
```diff
@@ -0,0 +1,54 @@
...
+                UserName = "User",
+                SenderName = "Sender",
+                UserData1 = 1,
+                UserData2 = 2
+            };
+            var emailContent = helper.RenderViewToString("EmailTemplate", model);
+            Console.WriteLine(emailContent);
+        }
+
+        private static void ConfigureDefaultServices(IServiceCollection services, string path)
```
rynowak: I think `path` here requires a little more fanfare for a reader to understand exactly what this does.

samples/Mvc.RenderViewToString/RazorViewHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+                        metadataProvider: new EmptyModelMetadataProvider(),
+                        modelState: new ModelStateDictionary())
+                    {
+                        Model = model
+                    },
+                    new TempDataDictionary(
+                        actionContext.HttpContext,
+                        _tempDataProvider),
+                    output,
+                    new HtmlHelperOptions());
```
rynowak: Interestingly, I don't see a custom RazorHost or base class for the view here, so what intellisense do you get? Does it include HtmlHelper and UrlHelper but those things don't work?

samples/Mvc.RenderViewToString/Program.cs
dougbu: @rynowak is the pattern you'd like to highlight the following? Or do you have a different suggestion?
``` c#
  services.AddSingleton<RazorViewHelper>();
  ...
  var helper = provider.GetRequiredService<RazorViewHelper>();
```

samples/Mvc.RenderViewToString/EmailGraphHelper.cs
ajaybhargavb: This is simply present to demonstrate `@inject`.

samples/Mvc.RenderViewToString/RazorViewHelper.cs
rynowak: This should throw, easier to understand in a sample context.

samples/Mvc.RenderViewToString/RazorViewHelper.cs
ajaybhargavb: That is correct. Those things don't work. Spoke to @Eilon about UrlHelper and he suggested just throwing an exception when no Router is available.

samples/Mvc.RenderViewToString/RazorViewHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace Mvc.RenderViewToString
+{
+    public class RazorViewHelper
+    {
+        private ICompositeViewEngine _viewEngine;
+        private ITempDataProvider _tempDataProvider;
+        private IServiceProvider _serviceProvider;
+
+        public RazorViewHelper(
+            ICompositeViewEngine viewEngine,
```
pranavkm: Should this be `IRazorViewEngine`?

ajaybhargavb:
Updated.

samples/Mvc.RenderViewToString/Views/EmailPartial.cshtml
ajaybhargavb: It would. But we already use partial views here and adding a view component as well feels unnecessary.

Eilon: Yeah perhaps no urgency on adding much more to this sample just yet. The goal right now is to get something functional out the door, let the masses know (it's one of the most popular requests for MVC/Razor), and see what people think.

samples/Mvc.RenderViewToString/EmailGraphHelper.cs
dougbu: > This is simply present to demonstrate `@inject`.

I'm struggling with this because views make extensive use of the `@inject` infrastructure without users entering the keyword. Suggest the sample would be clearer without this.

If it stays, would prefer new names for this and `RazorViewHelper`.

samples/Mvc.RenderViewToString/Program.cs
dougbu: Ping @rynowak

dougbu:
:watch:

ajaybhargavb:
Updated.

dougbu:
:shipit:

PR #87, opened 2016-01-20
Migrate to using System.Runtime.Loader.AssemblyLoadContext

pakrym:
Or this should be rewritten somehow?
aspnet/dnx#3227
@davidfowl


davidfowl:
I'd rewrite it using the new API

pakrym:
AssemblyResolve? Or setting AssemblyLoadContext?

davidfowl:
Hmm they haven't merged AssemblyResolve as yet but it makes sense to show both patterns so we can see just how hard it actually is to implement.

PR #93, opened 2016-02-04
Set the server manually

hishamco:


Eilon:
What is the value in this change? The current code is fine as it is... in fact it's probably not a good idea to change *every* repo as it's good to have a variety of different ways to set things up. Please consult with us first before sending so many similar PRs to every repo.

hishamco:
Ok @Eilon, i'm so sorry for doing such thing.

Eilon:
No worries, but I think we could sometimes help each other save some time by discussing broad changes like this first.

hishamco:
Thanks so much :smile:

Eilon:
Re-opening this, but for a different reason. I talked to @JunTaoLuo and we need this change because with dotnet-cli the hosting.json file by default won't get published, so we have two ways to fix this: 1. Remove the file and do the configuration in code; 2. Change the hosting.json file to be published to the target folder.

In most cases (1) is easier, so we'll do that. There might be some cases in which we need publishing to happen anyway, so only (2) would be an option.

GuardRex:
@Eilon Thank you for going with option (1). :+1:

PR #101, opened 2016-02-24
Add Localization ConfigurationRequestCultureProvider

hishamco:
/cc @Eilon  @ryanbrandenburg

### DIFF
--- a/Entropy.sln
+++ b/Entropy.sln
@@ -107,6 +107,8 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Config.CustomProviders.Web"
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Localization.EntityFramework", "samples\Localization.EntityFramework\Localization.EntityFramework.xproj", "{B456386A-D9E6-421B-9ACB-AFEE03945331}"
 EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Localization.ConfigurationRequestCultureProvider", "samples\Localization.ConfigurationRequestCultureProvider\Localization.ConfigurationRequestCultureProvider.xproj", "{0162B7C5-196C-4513-9115-F5E2958526CD}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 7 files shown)


Eilon:
This sample looks like it's better served by just a blog post, no? Ultimately there are only 2-3 lines of code that seem especially interesting in this sample.

hishamco:
I already blogged about it in [Code Project](http://www.codeproject.com/Articles/1081513/Localization-Extensibility-in-ASP-NET-Core) as well as in [in my website](http://هشام.com/Articles/Details/4-%D9%86%D9%82%D8%A7%D8%B7-%D8%A7%D9%84%D8%AA%D9%85%D8%AF%D8%AF-%D9%88%D8%A7%D9%84%D8%AA%D8%B7%D9%88%D9%8A%D8%B1-%D9%81%D9%8A-%D8%A7%D9%84%D8%AA%D8%B9%D8%B1%D9%8A%D8%A8-%D9%81%D9%8A-%D8%A7%D9%84%D9%80-ASP.NET-5) (Arabic Version) which I talked about two way to extend the localization, while we already have an EntityFramework Localization provider, I think it 'll be nice to have another sample to show localization extensibility from different angle :smile:

Eilon:
I think this particular sample also isn't showing some best practices. E.g. it's loading, parsing, and reading config on every single request.

hishamco:
FYI the sample read the config data on each request, so if you trying to change the cultures while the application is running it will reflect on the sample.
What I'm doing here is something like **Config.CustomProviders.Web** showing how you can use custom provider to read the culture infos from, if you have any improvements that I can made to the sample I will do, but I implement the core things is needed

Eilon:
The config system has built-in support for reloading while the app is running. See the Logging repo for example on how loggers use re-loaded config while the app is running. (Search for the word `reload` in that repo.)

But aside from that, I feel that this example is really just showing how to implement a fairly simple interface, so I'm not sure I see a lot of value there. Each sample we add includes a maintenance cost so we need to be a bit careful about what we add.

ryanbrandenburg:
I'm with @Eilon on this one, this is a good project to have out there for people to consult, but I don't think it's really core enough to enough people's use cases to include as a sample.

hishamco:
Thanks @Eilon, what I have seen here either make this sample rich enough or I will implement another rich sample for example **DomainRequestCultureProvider** which gets the culture infos from the domain.

Also there's a humble idea to such provider into **LocalizationSample**, because it's a rich enough example also the people can find a simple ```RequestCultureProvider```
Your thought @Eilon

Eilon:
I'm closing this for now. BTW the goal of Entropy isn't to have every possible sample. They're typically based off of common requests and scenarios. As such, please don't just make this sample richer because that still doesn't necessarily mean it's a good fit for this repo.

hishamco:
I understand, thanks a lot :smile:

PR #103, opened 2016-02-25
Don't run functional tests in parallel

pranavkm:


dougbu:
:shipit: w/ same CI success proviso as for aspnet/Localization#219.

PR #106, opened 2016-03-08
Fixed CustomRoutingConvention Test

kichalla:
@rynowak I am going to send one more PR in MVC for the exception `value cannot be null`


rynowak:
:shipit:

PR #109, opened 2016-03-21
Adding custom ConfigurationProvider to parse *.config

conniey:
This adds a parser that can parse *.config files. An example of the types of content it can parse is in the `Web.config`


samples/Config.CustomConfigurationProviders.Web/Extensions/ConfigurationProviderKeyExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public static class ConfigurationProviderKeyExtensions
+    {
+        /// <summary>
+        /// Gets the full key to search for in an IConfigurationRoot
+        /// </summary>
+        /// <example>
+        /// Given the configuration file:
+        /// <configuration>
+        ///     <nodea>
```
Eilon: Should check if this doc comment actually works. I'm guessing you need to XML-encode it (basically same as HTML-encode), otherwise none of these tags will show up.

samples/Config.CustomConfigurationProviders.Web/Extensions/ConfigurationProviderKeyExtensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Extensions.Configuration.Extensions
```
Eilon: I recommend against having *another* `.Extensions` in the namespace. Name this namespace after whatever it is about.

samples/Config.CustomConfigurationProviders.Web/Extensions/ConfigurationProviderKeyExtensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
Eilon: Use the exact license header that's used elsewhere in this repo (though admittedly many files are missing it entirely):

https://github.com/aspnet/Entropy/blob/dac7c2f64567956f3a6eafb24b51f59f62b0b3b3/samples/Mvc.Modules/ModuleFramework/IAsyncModuleFilter.cs#L1-L2

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+            if (!_loadFromFile)
+            {
+                document = XDocument.Parse(_configuration);
+                return true;
+            }
+
+            const int MAX_ATTEMPTS = 5;
+            int attempts = 0;
+
+            while (attempts < MAX_ATTEMPTS)
```
Eilon: Does this really need retry logic? No other providers have this.

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                try
+                {
+                    document = XDocument.Load(_configuration);
+                    return true;
+                }
+                catch (IOException e)
+                {
+                    Trace.TraceError($"Unable to open {_configuration}! " + Environment.NewLine + $"Exception: {e}");
```
Eilon: If you need to log, use `ILogger`.

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProviderExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        /// <summary>
+        /// Adds configuration values of a *.config file to the ConfigurationBuilder
+        /// </summary>
+        /// <param name="builder">Builder to add configuration values to</param>
+        /// <param name="configContents">Contents of *.config file</param>
+        /// <param name="parsers">Additional parsers to use to parse the config contents</param>
+        public static IConfigurationBuilder AddConfiguration(this IConfigurationBuilder builder, string configContents, params IConfigurationParser[] parsers)
+        {
+            if (string.IsNullOrEmpty(configContents))
```
Eilon: The check is for null-or-empty and the exception is for null.

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigurationAction.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// This represents the action for that *.config element.
+    /// THere are 3 possible element names:
+    ///     1. add
+    ///     2. remove
+    ///     3. clear
+    /// https://msdn.microsoft.com/en-us/library/aa903313(v=vs.71).aspx
+    /// </summary>
+    internal enum ConfigurationAction
+    {
+        add,
```
Eilon: Recommend following common naming conventions and PascalCasing these.

samples/Config.CustomConfigurationProviders.Web/project.json
```diff
@@ -0,0 +1,46 @@
...
+  "compilationOptions": {
+    "emitEntryPoint": true
+  },
+
+  "content": [
+    "Web*.config"
+  ],
+
+  "dependencies": {
+    "Microsoft.Extensions.Configuration": "1.0.0-rc1-final",
```
Eilon: Should probably use `*`'s in these, no?

conniey: I had it as * before, but `Microsoft.Extensions.Configuration.Constants.KeyDelimiter` doesn't exist in the rc2 packages that are getting restored. :( Is that something that is being removed?

Eilon: I think tis is what you're looking for: https://github.com/aspnet/Configuration/blob/70b9b0325be999e80bc97dfda095b73e0e17c790/src/Microsoft.Extensions.Configuration.Abstractions/ConfigurationPath.cs#L14

The key point is that you can't mix versions of packages (e.g. RC1 and RC2), they always have to be a "coherent set," so either everything needs to be hard-coded to a specific version, or everything needs to float together. In the case of the repos here we never hard-code the version because we always want it to compile against the most recent builds.

conniey: > The key point is that you can't mix versions of packages (e.g. RC1 and RC2), they always have to be a "coherent set,"
Oh, I see I see. Thank you for teaching me something new! I will fix that.

samples/Config.CustomConfigurationProviders.Web/Extensions/ConfigurationProviderKeyExtensions.cs
conniey: Fixed

samples/Config.CustomConfigurationProviders.Web/Extensions/ConfigurationProviderKeyExtensions.cs
conniey: I changed it to `Utilities` but am open to any other suggestions

samples/Config.CustomConfigurationProviders.Web/Extensions/ConfigurationProviderKeyExtensions.cs
conniey: Fixed. I escaped all of the XML strings.

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
conniey: Fixed

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
conniey: Fixed.

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProviderExtensions.cs
conniey: I split it into two checks. One for null and the other for empty.

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigurationAction.cs
conniey: Fixed

conniey:
@Eilon : Thanks for reviewing my PR. I fixed all of the issues you pointed out. Is there anything else? :)

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProviderExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// Adds configuration values of a *.config file to the ConfigurationBuilder
+        /// </summary>
+        /// <param name="builder">Builder to add configuration values to</param>
+        /// <param name="configContents">Contents of *.config file</param>
+        /// <param name="parsers">Additional parsers to use to parse the config contents</param>
+        public static IConfigurationBuilder AddConfiguration(this IConfigurationBuilder builder, string configContents, params IConfigurationParser[] parsers)
+        {
+            if (configContents == null)
+            {
+                throw new ArgumentNullException(nameof(configContents), "Contents for configuration cannot be null!");
```
Eilon: ArgNull doesn't really need a message.

samples/Config.CustomConfigurationProviders.Web/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProviderExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// </summary>
+        /// <param name="builder">Builder to add configuration values to</param>
+        /// <param name="path">Path to *.config file</param>
+        /// <param name="optional">true if file is optional; false otherwise</param>
+        /// <param name="parsers">Additional parsers to use to parse the config file</param>
+        public static IConfigurationBuilder AddConfigFile(this IConfigurationBuilder builder, string path, bool optional, params IConfigurationParser[] parsers)
+        {
+            if (string.IsNullOrEmpty(path))
+            {
+                throw new ArgumentNullException(nameof(path), "Path for configuration cannot be null/empty!");
```
Eilon: Exception type mismatch.

samples/Config.CustomConfigurationProviders.Web/project.json
```diff
@@ -0,0 +1,46 @@
+﻿{
+  "version": "1.0.0-*",
+  "description": "*.config ConfigurationProviders Library",
+  "authors": [ "Microsoft" ],
```
Eilon: Don't need this metadata: authors, tags, projectUrl, licenseUrl.

samples/Config.CustomConfigurationProviders.Web/Program.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.ConfigFile;
+using System;
+
+namespace Config.CustomConfigurationProviders.Web
```
Eilon: This project isn't actually a web site, so it probably shouldn't be named `*.Web`.

Perhaps the config provider project should be `Config.ConfigFileProvider`, and this sample app should be `Config.ConfigFileProvider.Sample`?

samples/Config.CustomConfigurationProviders.Web/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+        public static void Main()
+        {
+            var builder = new ConfigurationBuilder()
+                .AddConfigFile("Web.config");
+            var configuration = builder.Build();
+            var configurationManager = new ConfigurationManager(configuration);
+
+            Console.WriteLine("---------- AppSettings ----------");
+            foreach (var kvp in configurationManager.AppSettings)
+                Console.WriteLine($"{kvp.Key}: [{kvp.Value}]");
```
Eilon: Wrap these statements in `{ ... }` (spread out over several lines).

samples/Config.CustomConfigurationProviders.Web/Utilities/ConfigurationProviderKeyExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        ///             &lt;remove key=&quot;SomeTestKey&quot; /&gt;
+        ///         &lt;/nested&gt;
+        ///     &lt;/nodeb&gt;
+        /// &lt;/configuration&gt;
+        /// To get the key for NestedKey2:
+        ///     var key = GetFullKey(&quot;nodeb&quot;, &quot;nested&quot;, &quot;NestedKey2&quot;);
+        /// it would return something like: &quot;nodeb:nested:NestedKey2&quot;
+        /// &lt;/example&gt;
+        /// </example>
+        public static string GetFullKey(params string[] values)
```
Eilon: Is this used anywhere?

samples/Config.CustomConfigurationProviders.Web/Utilities/ConfigurationSectionExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Configuration.Utilities
+{
+    /// <summary>
+    /// Helper functions to get strongly typed values from a IConfigurationSection
+    /// </summary>
+    public static class ConfigurationSectionExtensions
```
Eilon: Are these used anywhere?

Eilon:
:clock1:

conniey:
@Eilon I fixed all of your comments

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            XDocument document;
+
+            if (!TryLoadXDocument(out document))
+            {
+                return;
+            }
+
+            Data = new SortedDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
Eilon: Is there a reason to use a sorted dictionary for these? I would think that file-order is fine to preserve.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            document = null;
+            return false;
+        }
+
+        /// <summary>
+        /// Given an XElement tries to parse that element using any of the KeyValueParsers
+        /// and adds it to the results dictionary
+        /// </summary>
+        private void ParseElement(XElement element, ref Stack<string> context, ref SortedDictionary<string, string> results)
```
Eilon: Why is the `results` parameter `ref`?

Eilon: And same for `context`?

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+            parsersToUse.AddRange(parsers);
+
+            _parsers = parsersToUse.ToArray();
+        }
+
+        public override void Load()
+        {
+            if (_loadFromFile && !_isOptional && !File.Exists(_configuration))
+            {
+                throw new FileNotFoundException("Could not find configuration file to load!", _configuration);
```
Eilon: Exception messages and log messages probably don't need exclamation points.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
Eilon: Also, the `Data` property is set here to an empty dictionary, and then a few lines below set again to the real dictionary with actual data. Can probably skip this line here.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+                document = XDocument.Parse(_configuration);
+                return true;
+            }
+
+            try
+            {
+                document = XDocument.Load(_configuration);
+                return true;
+            }
+            catch (IOException e)
```
Eilon: I wouldn't catch this exception at all. The problem with catching it here is that if the user doesn't have logging enabled, they'll never know something went wrong. However, this exception really is a fatal exception - the app *must not start* if the configuration data can't be loaded. As such, just let the exception bubble up and let the app die. ASP.NET Core (or whoever the caller is) will deal with the exception and logging on its own.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProviderExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <param name="parsers">Additional parsers to use to parse the config contents</param>
+        public static IConfigurationBuilder AddConfiguration(this IConfigurationBuilder builder, string configContents, params IConfigurationParser[] parsers)
+        {
+            if (configContents == null)
+            {
+                throw new ArgumentNullException(nameof(configContents));
+            }
+            else if (string.IsNullOrEmpty(configContents))
+            {
+                throw new ArgumentException(nameof(configContents), "Contents for configuration cannot be empty!");
```
Eilon: The order of these two parameters is incorrect (need to fix in a few places in this PR). https://msdn.microsoft.com/en-us/library/sxykka64(v=vs.110).aspx

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigurationManager.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        /// <summary>
+        /// Given a set of keys, will join them and get the value at that level
+        /// </summary>
+        /// <param name="keys">Names of the sections from top-level to lowest level</param>
+        /// <returns>The value of that key</returns>
+        public string GetValue(params string[] keys)
+        {
+            if (keys.Length == 0)
+                throw new ArgumentException("Need to provide keys", nameof(keys));
```
Eilon: Wrap in curly braces.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/IConfigurationParser.cs
```diff
@@ -0,0 +1,38 @@
...
+        ///       &lt;nestedSectionA&gt;
+        ///         &lt;add key=&quot;keyone&quot; value=&quot;value1&quot; /&gt;
+        ///       &lt;/nestedSectionA&gt;
+        ///     &lt;/sectionZ&gt;
+        ///   &lt;/configuration&gt;
+        /// while parsing the element &lt;add key=&quot;keyone&quot; value=&quot;value1&quot; /&gt;, the current elements on the stack could be:
+        ///  1. &quot;nestedSectionA&quot;
+        ///  2. &quot;sectionZ&quot;
+        /// </example>
+        void ParseElement(XElement element, ref Stack<string> context, ref SortedDictionary<string, string> results);
```
Eilon: Also should check why these parameters are `ref`.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/KeyValueParser.cs
```diff
@@ -0,0 +1,178 @@
...
+                    break;
+                case ConfigurationAction.Remove:
+                    results.Remove(fullkey);
+                    break;
+                default:
+                    throw new NotSupportedException($"Unsupported action! [{action}]");
+            }
+        }
+
+        private string GetKey(Stack<string> context, string name)
```
Eilon: Make this `static`?

samples/Config.CustomConfigurationProviders.Sample/project.json
```diff
@@ -0,0 +1,43 @@
+﻿{
+  "version": "1.0.0-*",
+  "description": "*.config ConfigurationProviders Library",
```
Eilon: This sample app doesn't need a `description`.

Eilon:
Looking good, a few more comments :watch:

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+                return;
+            }
+
+            Data = new SortedDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+            var context = new Stack<string>();
+            var dictionary = new SortedDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var child in document.Root.Elements())
+            {
+                ParseElement(child, ref context, ref dictionary);
```
mjrousos: Why do we need the dictionary local? Why not just pass Data to ParseElement directly?

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
conniey: I was trying to be explicit that those objects were being modified for each call. But its easy to remove.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
conniey: Removed

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
conniey: That makes sense. Removed.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProviderExtensions.cs
conniey: Good catch. I fixed all those.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/IConfigurationParser.cs
conniey: See [comment](https://github.com/aspnet/Entropy/pull/109#discussion_r57533007) above.

conniey:
@Eilon  Thanks for taking another look at it. :D I fixed all those comments.

samples/Config.CustomConfigurationProviders.Sample/project.json
```diff
@@ -0,0 +1,42 @@
...
+  "dependencies": {
+    "Microsoft.Extensions.Configuration": "1.0.0-*",
+    "Microsoft.Extensions.Configuration.FileExtensions": "1.0.0-*",
+    "Microsoft.Extensions.Logging.Abstractions": "1.0.0-*",
+    "System.Collections.Immutable": "1.2.0-*",
+    "System.Xml.XDocument": "4.0.11-*"
+  },
+
+  "commands": {
+    "Config.CustomConfigurationProviders.Web": "Config.CustomConfigurationProviders.Web"
```
Eilon: Fix command name.

Eilon:
Looking very good, just one last comment, then we can merge.

BTW when making additional changes as a result of code reviews, I recommend just pushing *additional* commits - that way reviewers can see what changed without having to re-review the whole change. When the existing commits are amended, there's no way to know what's new. Then, once the whole thing is signed off, you can combine all the commits into one by "squashing" them, and that keeps the history in the repo nice and clean (no point in seeing dozens of "fixed code review comments" commits once it's all done).

samples/Config.CustomConfigurationProviders.Sample/project.json
conniey: Fixed

conniey:
> BTW when making additional changes as a result of code reviews, I recommend just pushing additional commits - that way reviewers can see what changed without having to re-review the whole change. When the existing commits are amended, there's no way to know what's new. Then, once the whole thing is signed off, you can combine all the commits into one by "squashing" them, and that keeps the history in the repo nice and clean (no point in seeing dozens of "fixed code review comments" commits once it's all done).

Thanks! I learned something new today. Going forwards, I'll do that to make it easier to follow. :)

Eilon:
This is :shipit: from me. Looping in @divega @HaoK because they're the config experts, so I at least want them to see it. If they sign off, we will merge.

samples/Config.CustomConfigurationProviders.Sample/Microsoft.Extensions.Configuration.ConfigFile/ConfigFileConfigurationProvider.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.IO;
+using System.Xml.Linq;
+
+namespace Microsoft.Extensions.Configuration.ConfigFile
+{
+    /// <summary>
+    /// This ConfigurationProvider tries to parse .config files from System.Configuration
+    /// https://msdn.microsoft.com/en-us/library/system.configuration.configuration(v=vs.110).aspx
+    /// It can process configurations in the form: 
+    ///     1. &lt;add key=&quot;&quot; value=&quot;&quot; /&gt;
```
muratg: Thanks for writing doc comments!  Though, these do not render properly in the IDE. Give it a shot by hovering over an instance with your mouse pointer.  For example, lists like above need to be formatted as described here: https://msdn.microsoft.com/en-us/library/y3ww3c7e.aspx

conniey:
@muratg : Thanks for that tip! I fixed those XML lists.

divega:
In general it looks good as a sample to me (although I haven't tried to dig into the fine details). There are a couple of upcoming changes that will have an impact on this:

- https://github.com/aspnet/Configuration/pull/407 introduces a changes in the patterns around the `ConfigurationBuilder` that will likely break this sample.

- http://github.com/aspnet/Configuration/pull/414 introduces an extension method `GetConnectionString()` that works assuming a configuration section called "ConnectionStrings". Not sure we would do the same kind of thing for "AppSettings" (besides .config files the way Azure Web Sites configuration works could also make that somewhat compelling) but it makes me wonder about the value of the `ConfigurationManager` class. It seems to add an extra API layer that looks a bit more backwards compatible but I am not sure it is really fundamentally different from what you can get with a couple of extension methods on `IConfiguration`.

conniey:
@divega 
> * aspnet/Configuration#407 introduces a changes in the patterns around the ConfigurationBuilder that will likely break this sample.
Should I wait for this change to go in and then revamp my sample?

>* aspnet/Configuration#414 introduces an extension method GetConnectionString()...
I removed the ConfigurationManager class and am using static methods now. Is that OK?

divega:
Looks good. 

> Should I wait for this change to go in and then revamp my sample?

Either way I think it is fine, but I believe both changes will be in within a week, so it may be better to wait.

JunTaoLuo:
Sample merged in https://github.com/aspnet/Entropy/commit/5d246b0f5ace7471a6f88b3c645282ae25d017f3 with some additional updates and cleanup. Thanks @conniey!

PR #112, opened 2016-03-31
Custom middleware sample (handling SOAP requests)

mjrousos:


Eilon:
Closing because we are not planning to add this code to the repo. The Entropy repo is more about some simple sample apps that demonstrate a few key features so that the team can make sure everything is running smoothly.

PR #115, opened 2016-04-02
React to removal of IAssemblyProvider

ryanbrandenburg:
Needed to fix Universe, will merge right away, but comment if you have any gripes.

CC @javiercn


javiercn:
@rynowak We want to get rid of this sample, don't we?

javiercn:
@rynowak We wanted to get rid of this sample, ddn't we?

Issue #117, opened 2016-04-14
Add functional test for each sample


JunTaoLuo:
Although we build each sample, they may not run. This happens very often due to breaking changes and we don't discover this until the samples are tested manually. For example, I have fixed `Config.CustomProviders.Web` twice in https://github.com/aspnet/Entropy/commit/7e8cedf480f9061965082b2098087b27adce6379 and https://github.com/aspnet/Entropy/commit/4b56220383348073c2ee07dcad54b0635ee59901.

Currently our functional tests do not cover all the samples but we should at least have functional tests that launch each sample so regressions are caught more frequently.

JunTaoLuo:
Also consider expanding test coverage for IISExpress

JunTaoLuo:
Currently the following samples do not have associated tests:

- [ ] Config.ChangeNotification.Web
- [ ] Config.CustomConfigurationProviders.Sample
- [ ] Config.Providers.Web
- [ ] Content.Upload.Files
- [ ] Content.Upload.Forms
- [ ] Content.Upload.Multipart
- [ ] Data.InMemory
- [ ] Data.SqlServer
- [ ] Localization.EntityFramework
- [ ] Microsoft.AspNetCore.Mvc.TagHelpers.Localization.EntityFramework
- [ ] Logging.Elm
- [ ] MusicStore.Spa
- [ ] Mvc.FileUpload
- [ ] Mvc.Modules
- [ ] Mvc.RenderViewToString
- [ ] Mvc.UrlHelperSample.Web
- [ ] Owin.ConsoleApp.HelloWorld
- [ ] Owin.HelloWorld
- [ ] Owin.IAppBuilderBridge
- [ ] Owin.Nowin.HellWorld
- [ ] Owin.Nowin.WebSockets
- [ ] Owin.OwinWebSockets
- [ ] Project.CompilerSettings
- [ ] Project.Dependencies
- [ ] Project.ProjectReference
- [ ] Project.SharedFiles
- [ ] Project.WildcardSources
- [ ] Runtime.ApplicationEnvironment

Note that  testing some projects probably won't be tested. For example, Config.ChangeNotification.Web contains no code at all.

PR #118, opened 2016-04-14
Use UseServer() instead of UseKestrel() to allow override in tests

JunTaoLuo:
Only fixing the samples that are tested for now. We'll need to also update other samples when we add tests in https://github.com/aspnet/Entropy/issues/117. Note the localization failures are due to https://github.com/dotnet/cli/issues/467 which has been fixed by @anurse.

Review: @Tratcher @mikeharder 
Approval: @Eilon @DamianEdwards

### DIFF
--- a/samples/Builder.Filtering.Web/Startup.cs
+++ b/samples/Builder.Filtering.Web/Startup.cs
@@ -47,8 +47,10 @@ namespace Builder.Filtering.Web
         public static void Main(string[] args)
         {
             var host = new WebHostBuilder()
+                // We set the server by name before default args so that command line arguments can override it.
+                // This is used to allow deployers to choose the server for testing.
+                .UseServer("Microsoft.AspNetCore.Server.Kestrel")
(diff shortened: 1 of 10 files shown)


samples/Builder.Filtering.Web/Startup.cs
```diff
@@ -47,8 +47,10 @@ public void Configure(IApplicationBuilder app)
         public static void Main(string[] args)
         {
             var host = new WebHostBuilder()
+                // We set the server by name before default args so that command line arguments can override it.
+                // This is used to allow deployers to choose the server for testing.
+                .UseServer("Microsoft.AspNetCore.Server.Kestrel")
```
mikeharder: Should we use an expression like `typeof(KestrelServer).GetTypeInfo().Assembly.FullName` instead of hardcoding the string `"Microsoft.AspNetCore.Server.Kestrel"`?  The latter is more fragile if we ever rename the Kestrel assembly.

Tratcher:
:shipit:

PR #121, opened 2016-04-20
Remove npm \ gulp from publish

pranavkm:


JunTaoLuo:
Are we removing this from all of our samples? If so, you might want to check https://github.com/search?p=5&q=user%3Aaspnet+filename%3Aproject.json+npm&ref=searchresults&type=Code&utf8=%E2%9C%93. For example Identity has a sample that uses gulp/npm

pranavkm:
Filed an issue for Identity - https://github.com/aspnet/Identity/issues/804. The remainder are still in use so they look ok. This one is causing the build to fail.

JunTaoLuo:
:shipit: if it fixes the build.

blowdart:
Feel free to fix this in identity yourself :p

PR #123, opened 2016-04-27
Added new sample Mvc.FileUpload

kichalla:
@rynowak @Tratcher


samples/Config.CustomConfigurationProviders.Sample/Config.CustomConfigurationProviders.Sample.xproj
```diff
@@ -9,7 +9,7 @@
     <ProjectGuid>c2c08dd5-22ea-48dd-bda4-588dcd38048f</ProjectGuid>
     <RootNamespace>Microsoft.Extensions.Configuration.ConfigFile</RootNamespace>
     <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">..\..\artifacts\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
-    <OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\bin\$(MSBuildProjectName)\</OutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\</OutputPath>
```
kichalla: This was changed by the latest (the last recommended) tooling. I will revert it.

rynowak: Please don't fight the tools. It's going to do this every time you open the solution

samples/Mvc.RestrictiveFileUpload/RequestFormSizeLimitAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class RequestFormSizeLimitAttribute : Attribute, IAuthorizationFilter, IOrderedFilter
+    {
+        private readonly FormOptions _formOptions;
+
+        public RequestFormSizeLimitAttribute(long bufferBodyLengthLimitFormSizeInBytes, int order = 0)
+        {
+            _formOptions = new FormOptions()
+            {
+                BufferBodyLengthLimit = bufferBodyLengthLimitFormSizeInBytes
```
Tratcher: Don't invent new names. If you don't like the existing names file a bug against the original.

Tratcher: This doesn't actually work unless you also enable buffering.

samples/Mvc.RestrictiveFileUpload/Controllers/UsersController.cs
```diff
@@ -0,0 +1,37 @@
...
+        [HttpGet]
+        public IActionResult Create()
+        {
+            return View();
+        }
+
+        // Set the request form size limits *before* the antiforgery token validation filter is executed so that the
+        // limits are honored when the validation filter tries to read the form.
+        [HttpPost]
+        [RequestFormSizeLimit(bufferBodyLengthLimitFormSizeInBytes: 1024 * 1024, order: 1)]
```
Tratcher: This lowers the limit from the default. We don't expect that to be the common scenario. More likely people are going to start with low limits and then raise them on an as-needed basis. You can leave this as is, but lower the default limits in Startup.ConfigureServices.

samples/Mvc.RestrictiveFileUpload/Program.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public class Program
+    {
+        // Entry point for the application.
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
+                .UseKestrel()
+                .UseContentRoot(Directory.GetCurrentDirectory())
+                .UseDefaultHostingConfiguration(args)
```
Tratcher: Note UseDefaultHostingConfiguration was just removed from the entropy samples.

samples/Mvc.RestrictiveFileUpload/RequestFormSizeLimitAttribute.cs
Tratcher: Have you tested with files that exceed the limit?

samples/Mvc.RestrictiveFileUpload/project.json
```diff
@@ -0,0 +1,50 @@
...
+    "netcoreapp1.0": {
+      "imports": [
+        "dotnet5.6",
+        "dnxcore50",
+        "portable-net45+win8"
+      ]
+    }
+  },
+
+  "tools": {
```
Tratcher: your tools and scripts are out of date. See https://github.com/aspnet/ServerTests/blob/dev/test/ServerComparison.TestSites/project.json @moozzyk

samples/Mvc.RestrictiveFileUpload/web.Debug.config
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Tratcher: delete debug.config and release.config. The tooling only generated them by mistake.

samples/Mvc.RestrictiveFileUpload/RequestFormSizeLimitAttribute.cs
kichalla: Hence the design PR :-)...actually I am not very clear how this buffer limit stuff is supposed to work....(as i was mentioning offline), like when/what scenario would i want to explicitly buffer the request...

samples/Mvc.RestrictiveFileUpload/web.Release.config
```diff
@@ -0,0 +1,31 @@
...
+    finds an attribute "name" that has a value of "MyDB".
+    
+    <connectionStrings>
+      <add name="MyDB" 
+        connectionString="Data Source=ReleaseSQLServer;Initial Catalog=MyReleaseDB;Integrated Security=True" 
+        xdt:Transform="SetAttributes" xdt:Locator="Match(name)"/>
+    </connectionStrings>
+  -->
+  <system.web>
+    <compilation xdt:Transform="RemoveAttributes(debug)" />
```
moozzyk: we don't have support for xdt

samples/Mvc.RestrictiveFileUpload/project.json
```diff
@@ -0,0 +1,50 @@
...
+    "Microsoft.AspNetCore.Server.IISIntegration.Tools": "1.0.0-*"
+  },
+
+  "content": [
+    "wwwroot",
```
moozzyk: do you actually have anything in wwwroot?

kichalla:
:up: :date: 
There are 2 scenarios demonstrated in the sample. One for showing setting request form size limits on a granular basis (ex: at per-action level) and other one related to file upload.

samples/Mvc.FileUpload/appsettings.json
```diff
@@ -0,0 +1,10 @@
+﻿{
+  "Logging": {
+    "IncludeScopes": false,
+    "LogLevel": {
+      "Default": "Debug",
+      "System": "Information",
+      "Microsoft": "Debug"
```
kichalla: Will revert to the default `Information` level

samples/Mvc.FileUpload/MultipartRequestHelper.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.IO;
+using Microsoft.Net.Http.Headers;
+
+namespace Mvc.FileUpload
+{
+    public static class MultipartRequestHelper
```
kichalla: Borrowed from `FormFeature`

samples/Mvc.FileUpload/Filters/GenerateAntiforgeryTokenCookieForAjax.cs
```diff
@@ -0,0 +1,22 @@
...
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Mvc.FileUpload.Filters
+{
+    public class GenerateAntiforgeryTokenCookieForAjaxAttribute : ActionFilterAttribute
+    {
+        public override void OnActionExecuted(ActionExecutedContext context)
+        {
+            var antiforgery = context.HttpContext.RequestServices.GetService<IAntiforgery>();
```
kichalla: Could have used a service filter, but wanted to keep the sample simple

samples/Mvc.FileUpload/Controllers/FileController.cs
```diff
@@ -0,0 +1,130 @@
...
+        [DisableFormValueModelBinding]
+        [ValidateAntiForgeryToken]
+        public async Task<IActionResult> Upload()
+        {
+            if (!MultipartRequestHelper.IsMultipartContentType(Request.ContentType))
+            {
+                return BadRequest($"Expected a multipart request, but got '{Request.ContentType}'.");
+            }
+
+            //TODO: HOW CAN I MODEL BIND ALL THIS FORM URL ENCODED DATA LATER
```
kichalla: An open question: A common scenario we typically see is a user profile page where some text data is entered and a picture is uploaded. In this scenario how to bind the text data to a strongly typed object.

Tratcher: @rynowak this does seem to be a gap in the design.

samples/Mvc.FileUpload/Controllers/FileController.cs
```diff
@@ -0,0 +1,130 @@
...
+            return View();
+        }
+
+        // 1. We disable the form value model binding here to take control of handling potentially large files.
+        // 2. Typically antiforgery tokens are sent in request body, but since we do not want to read the request body
+        //    early, the tokens are made to be sent via headers. The antiforgery token filter first looks for tokens
+        //    in the request header and then falls back to reading the body.
+        [HttpPost]
+        [DisableFormValueModelBinding]
+        [ValidateAntiForgeryToken]
```
Tratcher: Could this specify the expected location rather than relying on built in ordering? If the token header was missing this would read the whole body, which is exactly what we're trying to avoid.

samples/Mvc.FileUpload/Controllers/FileController.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            if (!MultipartRequestHelper.IsMultipartContentType(Request.ContentType))
+            {
+                return BadRequest($"Expected a multipart request, but got '{Request.ContentType}'.");
+            }
+
+            //TODO: HOW CAN I MODEL BIND ALL THIS FORM URL ENCODED DATA LATER
+
+            // Used to accumulate all the form url encoded key value pairs in the request.
+            var formAccumulator = new KeyValueAccumulator();
```
Tratcher: Nothing in the sample consumes the accumulated values. Should there be a success view showing the fields and file stats?

samples/Mvc.FileUpload/MultipartRequestHelper.cs
Tratcher: You've had to borrow quite a bit of code here and in the controller. Are there more pieces of this that should be made public?

samples/Mvc.FileUpload/Startup.cs
```diff
@@ -0,0 +1,47 @@
...
+            Configuration = builder.Build();
+        }
+
+        public IConfigurationRoot Configuration { get; set; }
+
+        public void ConfigureServices(IServiceCollection services)
+        {
+            // Angular's default header name for sending the XSRF token.
+            services.AddAntiforgery(options => options.HeaderName = "X-XSRF-TOKEN");
+
```
Tratcher: Add app wide FormOptions to show that the controller attribute is overriding them.

samples/Mvc.FileUpload/Controllers/FileController.cs
kichalla: Sure.

samples/Mvc.FileUpload/Controllers/FileController.cs
rynowak: This isn't a problem if you want to buffer.

I think you can create an instance of `FormValueProvider` and use that to call try update model

samples/Mvc.FileUpload/Controllers/FileController.cs
```diff
@@ -0,0 +1,130 @@
...
+                ContentDispositionHeaderValue contentDisposition;
+                ContentDispositionHeaderValue.TryParse(section.ContentDisposition, out contentDisposition);
+
+                if (MultipartRequestHelper.HasFileContentDisposition(contentDisposition))
+                {
+                    var name = HeaderUtilities.RemoveQuotes(contentDisposition.Name) ?? string.Empty;
+                    var fileName = HeaderUtilities.RemoveQuotes(contentDisposition.FileName) ?? string.Empty;
+
+                    // Copy the uploaded file to local disk
+                    var targetFilePath = Path.Combine(_hostingEnvironment.ContentRootPath, Guid.NewGuid().ToString());
```
rynowak: Why would you put the uploaded files in the content root?

samples/Mvc.FileUpload/Controllers/FileController.cs
```diff
@@ -0,0 +1,130 @@
...
+                    // Do not limit the key name length here because the mulipart headers length
+                    // limit is already in effect.
+                    var key = HeaderUtilities.RemoveQuotes(contentDisposition.Name);
+                    MediaTypeHeaderValue mediaType;
+                    MediaTypeHeaderValue.TryParse(section.ContentType, out mediaType);
+                    var encoding = FilterEncoding(mediaType?.Encoding);
+                    using (var streamReader = new StreamReader(
+                        section.Body,
+                        encoding,
+                        detectEncodingFromByteOrderMarks: true,
```
rynowak: Should this have/support a BOM?

samples/Mvc.FileUpload/Filters/GenerateAntiforgeryTokenCookieForAjax.cs
```diff
@@ -0,0 +1,22 @@
...
+
+            // We can send the request token as a JavaScript-readable cookie, and Angular will use it by default.
+            var tokens = antiforgery.GetAndStoreTokens(context.HttpContext);
+            context.HttpContext.Response.Cookies.Append(
+                "XSRF-TOKEN",
+                tokens.RequestToken,
+                new CookieOptions() { HttpOnly = false });
```
rynowak: This is pretty cool as a filter, I like it.

samples/Mvc.FileUpload/MultipartRequestHelper.cs
rynowak: @Tratcher - this seems like this is really a lot of code to copy into a sample. Do we have a more recommended approash?

rynowak: Ah it wasn't showing your comments when I was looking at it. Yeah we need to improve this, this isn't practical.

samples/Mvc.FileUpload/Controllers/FileController.cs
Tratcher: If you're reading text it never hurts to filter out BOMs.

samples/Mvc.FileUpload/Filters/GenerateAntiforgeryTokenCookieForAjax.cs
kichalla: Question: An application could have a mix of pages where some use Ajax and some regular form post. Since this code sets a cookie which would be sent in subsequent requests to all pages of the app, I was initially wondering if doing a form post to a page would fail antiforgery validation (I say this because we recently made a change to look at the header first and then the form field). I tried testing few scenarios and the form post seems to succeed without any issue even though the value of the token is different between the header and the form field token. I am not sure if I am missing any scenarios. 
OR would it be safe to set the `CookieOptions.Path` to the url that we are posting to so that the scope of this cookie is limited only to that action.

samples/Mvc.FileUpload/Controllers/FileController.cs
kichalla: I wanted to keep the sample simple without much setup(example: setting up azure blog storage etc.) required. However, I think I could have a sample showing copying stream to a azure blob.

samples/Mvc.FileUpload/Controllers/FileController.cs
kichalla: @rynowak your thoughts?

kichalla: Filed: https://github.com/aspnet/Mvc/issues/4734

samples/Mvc.FileUpload/MultipartRequestHelper.cs
kichalla: I will file an issue in HttpAbstractions once I push this PR (want to reference the code that I wrote for this sample to give an idea)

kichalla:
:up: :date:

Tratcher:
:+1:

samples/Mvc.FileUpload/Controllers/FileController.cs
```diff
@@ -0,0 +1,174 @@
...
+                    //await cloudBlobContainer.CreateIfNotExistsAsync();
+
+                    //var blobId = Guid.NewGuid().ToString();
+                    //var cloudBlockBlob = cloudBlobContainer.GetBlockBlobReference(blobId);
+
+                    //// Copy the uploaded file's raw stream directly to the blob
+                    //await cloudBlockBlob.UploadFromStreamAsync(section.Body);
+
+                    //targetFilePath = blobId;
+                    //_logger.LogInformation($"Copied the uploaded file '{fileName}' to blob with id '{blobId}'.");
```
rynowak: I don't think there's any value to having this code here. This just bloats the sample - if we need to provide a sample for blob storage let's do that separately.

samples/Mvc.FileUpload/Controllers/FileController.cs
```diff
@@ -0,0 +1,174 @@
...
+        public async Task<IActionResult> Upload()
+        {
+            if (!MultipartRequestHelper.IsMultipartContentType(Request.ContentType))
+            {
+                return BadRequest($"Expected a multipart request, but got '{Request.ContentType}'.");
+            }
+
+            // Used to accumulate all the form url encoded key value pairs in the request.
+            var formAccumulator = new KeyValueAccumulator();
+            string targetFilePath = null;
```
rynowak: This can be moved into the block where it's used

samples/Mvc.FileUpload/MultipartRequestHelper.cs
kichalla: Filed: https://github.com/aspnet/HttpAbstractions/issues/638

rynowak:
Looks like https://github.com/aspnet/Entropy/pull/123/files#r64093232 still needs to be addressed

other than that :shipit: from me

samples/Mvc.FileUpload/Controllers/FileController.cs
```diff
@@ -0,0 +1,174 @@
...
+namespace Mvc.FileUpload.Controllers
+{
+    public class FileController : Controller
+    {
+        private readonly IHostingEnvironment _hostingEnvironment;
+        private readonly ILogger<FileController> _logger;
+
+        // Get the default form options so that we can use them to set the default limits for
+        // request body data
+        private static readonly FormOptions _defaultFormOptions = new FormOptions();
```
davidfowl: Why are you newing this up?

kichalla: In this sample request form is being parsed manually (in the sense that MultipartReader is being used to read individual parts and handled) compared to depending on FormFeature. Most of the code here is borrowed from FormFeature and I wanted to use the default limits here too.

kichalla:
:up: :date: @Tratcher I updated the PR with appwide FormOptions settings

Tratcher:
:shipit:

kichalla:
86147eda46a476b8acc409df0eaa9b0eca730e94

Issue #129, opened 2016-05-25
Localization.StarterWeb test failures


JunTaoLuo:
Still investigating but just going to list what I have found so far.

The failure I'm seeing on the CI is at http://aspnetci/viewLog.html?buildId=100911&buildTypeId=Lite_UniverseTest&tab=buildLog#_focus=64532 which is caused by:
```
info: Localization.StarterWeb:Kestrel:Clr:x64[0]
      System.InvalidOperationException: Can not find assembly file Microsoft.DotNet.Cli.Utils.dll at 'C:\Users\johluo\AppData\Local\Temp\058a9647-0c7a-4181-82f3-158ee5541928\refs,C:\Users\johluo\AppData\Local\Temp\058a9647-0c7a-4181-82f3-158ee5541928\'
info: Localization.StarterWeb:Kestrel:Clr:x64[0]
         at Microsoft.Extensions.DependencyModel.Resolution.AppBaseCompilationAssemblyResolver.TryResolveAssemblyPaths(CompilationLibrary library, List`1 assemblies)
info: Localization.StarterWeb:Kestrel:Clr:x64[0]
         at Microsoft.Extensions.DependencyModel.Resolution.CompositeCompilationAssemblyResolver.TryResolveAssemblyPaths(CompilationLibrary library, List`1 assemblies)
info: Localization.StarterWeb:Kestrel:Clr:x64[0]
         at Microsoft.Extensions.DependencyModel.CompilationLibrary.ResolveReferencePaths()
info: Localization.StarterWeb:Kestrel:Clr:x64[0]
         at Microsoft.AspNetCore.Mvc.Razor.Compilation.MetadataReferenceFeatureProvider.PopulateFeature(IEnumerable`1 parts, MetadataReferenceFeature feature)
info: Localization.StarterWeb:Kestrel:Clr:x64[0]
         at Microsoft.AspNetCore.Mvc.ApplicationParts.ApplicationPartManager.PopulateFeature[TFeature](TFeature feature)
```
This seems to have started after the changes in https://github.com/aspnet/Mvc/issues/4498. The reason this error occurs is that for the `net451 TFM`, while the assembly `Microsoft.DotNet.Cli.Utils.dll` appears in the build output of the project, it does not show up in the published output. What I'm curious about is why this dotnet cli assembly is being loaded even in full framework and whether this is a publish issue or an issue with how Razor is loading the assemblies. 

cc @cesarbs @pranavkm

[...]

PR #131, opened 2016-05-25
Temporary workaround for cli issue

JunTaoLuo:
Workaround for #129. This should resolve the test failures we are seeing on the CI until cli issue is fixed. @cesarbs @troydai 

cc @pakrym


cesarbs:
:ship: Do we have an issue to keep track of this so we can revert it later when the CLI issue is fixed?

JunTaoLuo:
we'll use the same issue.

PR #133, opened 2016-05-29
React to aspnet/Razor#705

pranavkm:
cc @NTaylorMullen \ @dougbu


test/FunctionalTests/MvcTests/TagHelperSampleTest.cs
```diff
@@ -232,24 +232,24 @@ private static void AssertRedirectsToHome(HttpResponseMessage response)
             return content;
         }
 
-        private static void AssertContent(string resourceFile, string actual)
+        private void AssertContent(string resourceFile, string actual)
```
dougbu: Why isn't this still `static`?

test/FunctionalTests/MvcTests/TagHelperSampleTest.cs
```diff
@@ -47,7 +47,7 @@ public TagHelperSampleTest(SampleTestFixture<TagHelperSample.Web.Startup> fixtur
             AssertRedirectsToHome(createBobbyResponse);
 
             var expectedMediaType = MediaTypeHeaderValue.Parse("text/html; charset=utf-8");
-            var resourceFile = "FunctionalTests.resources.TagHelperSample.Web.Home.Index.html";
+            var resourceFile = "TagHelperSample.Web.Home.Index.html";
```
dougbu: Why was a change necessary here? Appear to be calculating the same name that was previously provided.

test/FunctionalTests/MvcTests/SampleTestFixture.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
```
dougbu: Is this needed when `GENERATE_BASELINES` isn't set? Wondering about Remove and Sort Usings doing the Wrong Thing<sup>:tm:</sup>.

test/FunctionalTests/resources/TagHelperSample.Web.Home.Index-Reset.html
```diff
@@ -21,7 +21,7 @@
         document.write("<p>foo is from page</p>");
     }
 </script>
-<script>(window.foo||document.write("<script src=\"\/fallback.js\"><\/script>"));</script>
+<script>(window.foo||document.write("\u003Cscript src=\u0022\/fallback.js\u0022\u003E\u003C\/script\u003E"));</script>
```
dougbu: Finally got to the actual Razor reaction :smiley_cat:

dougbu:
:shipit: Since this is a build break, fell free to handle my comments in a follow-up.

test/FunctionalTests/MvcTests/SampleTestFixture.cs
pranavkm: I'll check.

test/FunctionalTests/MvcTests/TagHelperSampleTest.cs
pranavkm: It becomes trickier to figure out the file name when updating (we'd have to remove the dots until `resources.` with slashes). We pass in the resource file name now and construct the full resource location or the path based on what we're doing.

test/FunctionalTests/MvcTests/TagHelperSampleTest.cs
pranavkm: Yup. I'll revert this

PR #139, opened 2016-06-08
Fix MoviesController_Pages_ReturnSuccess test failure

pakrym:
Fixes NRE failure in `MoviesController_Pages_ReturnSuccess`, which was hidden by `UseDeveloperExceptionPage`.

### DIFF
--- a/samples/Mvc.TagHelperSample.Web/Services/ISignalTokenProviderService.cs
+++ b/samples/Mvc.TagHelperSample.Web/Services/ISignalTokenProviderService.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace TagHelperSample.Web.Services
+{
+    public interface ISignalTokenProviderService
+    {
+        IChangeToken GetToken(string key);
+
+        void SignalToken(string key);
+    }
+}
--- a/samples/Mvc.TagHelperSample.Web/Services/MoviesService.cs
+++ b/samples/Mvc.TagHelperSample.Web/Services/MoviesService.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Primitives;
 using TagHelperSample.Web.Models;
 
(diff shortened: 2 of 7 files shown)


pranavkm:
It would be nice to fix the pattern we have here and use https://github.com/aspnet/Mvc/blob/dev/test/WebSites/HtmlGenerationWebSite/SignalTokenProviderService.cs to manage the change tokens. (Particularly because this is a sample). Although that might require a bunch of changes.

pakrym:
@pranavkm I'm OK with copying it, but do we want to complicate TagHelperSample with it?

pranavkm:
It's a sample and we should try and make it as close to the pattern we expect users to use.

pranavkm:
cc @NTaylorMullen 

:shipit: run it by @Eilon for ask mode approval.

samples/Mvc.TagHelperSample.Web/Services/ISignalTokenProviderService.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace TagHelperSample.Web.Services
+{
+    public interface ISignalTokenProviderService<TKey>
```
NTaylorMullen: Doesn't need `TKey`. The implementation can assume the type. It's only ever used for `FeaturedMovies`. No need to make it more complex :smile:

samples/Mvc.TagHelperSample.Web/Services/MoviesService.cs
```diff
@@ -44,15 +55,21 @@ public string GetCriticsQuote(out IChangeToken expirationToken)
                 "Bravo!"
             };
 
-            expirationToken = new CancellationChangeToken(_quotesTokenSource.Token);
-            return quotes[_random.Next(0, quotes.Length)];
+            string quote;
+            if (!_memoryCache.TryGetValue(movieName, out quote))
```
NTaylorMullen: https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs#L98

samples/Mvc.TagHelperSample.Web/Services/MoviesService.cs
```diff
@@ -5,37 +5,48 @@
...
+            _memoryCache = memoryCache;
+            _tokenProviderService = tokenProviderService;
+        }
 
-            expirationToken = new CancellationChangeToken(_featuredMoviesTokenSource.Token);
-            return GetMovies().OrderBy(m => m.Rank).Take(2);
+        public IEnumerable<FeaturedMovies> GetFeaturedMovies()
+        {
+            IEnumerable<FeaturedMovies> movies;
+            if (!_memoryCache.TryGetValue(FeaturedTokenKey, out movies))
```
NTaylorMullen: https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Abstractions/MemoryCacheExtensions.cs#L98

samples/Mvc.TagHelperSample.Web/Services/SignalTokenProviderService.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.Extensions.Primitives;
+
+namespace TagHelperSample.Web.Services
+{
+    public class SignalTokenProviderService : ISignalTokenProviderService
+    {
+        private readonly ConcurrentDictionary<object, ChangeTokenInfo> _changeTokens
+            = new ConcurrentDictionary<object, ChangeTokenInfo>();
+
+        public IChangeToken GetToken(object key)
```
NTaylorMullen: We only ever pass in string keys, might be worth simplifying this to express that.

samples/Mvc.TagHelperSample.Web/Services/MoviesService.cs
pakrym:
```
 public IEnumerable<FeaturedMovies> GetFeaturedMovies()
        {
            var movies = _memoryCache.GetOrCreate(FeaturedTokenKey,
                (entry) =>
                {
                    var value = GetMovies().OrderBy(m => m.Rank).Take(2);

                    entry.SetValue(value).SetOptions(new MemoryCacheEntryOptions()
                        .AddExpirationToken(_tokenProviderService.GetToken(FeaturedTokenKey)));
                    return value;
                }
            );
            return movies;
        }
```

Looks bad

pakrym: Nevermind, did it better

pakrym:
@NTaylorMullen done

NTaylorMullen:
:shipit: after approval

PR #147, opened 2016-07-05
Fix MvcRenderViewToString and add a test to verify it works

pranavkm:


pranavkm:
cc @dougbu \ @ajaybhargavb

samples/Mvc.RenderViewToString/Program.cs
```diff
@@ -33,25 +42,38 @@ public static void Main()
...
+                WebRootFileProvider = fileProvider,
             });
             services.Configure<RazorViewEngineOptions>(options =>
             {
                 options.FileProviders.Clear();
-                options.FileProviders.Add(new PhysicalFileProvider(customApplicationBasePath ?? applicationEnvironment.ApplicationBasePath));
+                options.FileProviders.Add(fileProvider);
             });
             var diagnosticSource = new DiagnosticListener("Microsoft.AspNetCore");
+            services.AddSingleton<ObjectPoolProvider, DefaultObjectPoolProvider>();
```
pranavkm: @davidfowl - this is so utterly crummy.

samples/Mvc.RenderViewToString/project.json
```diff
@@ -1,7 +1,10 @@
 {
   "buildOptions": {
     "emitEntryPoint": true,
-    "preserveCompilationContext": true
+    "preserveCompilationContext": true,
+    "copyToOutput": {
+      "include": "Views"
```
pranavkm: We don't set the content root to the current working directory anywhere so this is required to make `dotnet run` work  out of the box.

test/FunctionalTests/MvcTests/RazorEmbeddedResource.cs
```diff
@@ -0,0 +1,55 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Reflection;
+using Xunit;
+
+namespace EntropyTests
+{
+    public static class RazorEmbeddedResource
```
dougbu: Any changes from the `TagHelperSampleTest` source?

samples/Mvc.RenderViewToString/Program.cs
dougbu: How would you prefer hosting-provided services be handled?

dougbu:
LGTM but I defer to @ajaybhargavb on the squirrel.

test/FunctionalTests/MvcTests/RazorEmbeddedResource.cs
pranavkm: The method was made public and it's referencing this type to locate the project, but nothing apart from that.

samples/Mvc.RenderViewToString/Program.cs
pranavkm: Since Razor depends on it and we sort of expect it to work in non-hosted scenarios, we should add it as part of `AddRazor` \ `AddMvc`. It's a no-op operation if it already exists via Hosting.

pranavkm:
Ping @ajaybhargavb

ajaybhargavb:
:shipit:

ajaybhargavb:
Does it make sense to add one more test with `customApplicationBasePath=null`?

pranavkm:
It's kinda tricky because you have to in some way modify the `ApplicationBasePath` to point to the test app when it's pointing at the test project. The end results are going to look pretty similar to passing in a path.

ajaybhargavb:
Got it

PR #152, opened 2016-08-01
ContentRoot Fix

shanselman:
This can't work, so I added the ContentRoot calls to make sure it runs.

### DIFF
--- a/samples/Localization.StarterWeb/Startup.cs
+++ b/samples/Localization.StarterWeb/Startup.cs
@@ -1,21 +1,21 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
+using Localization.StarterWeb.Models;
+using Localization.StarterWeb.Services;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
+using Microsoft.AspNetCore.Localization;
+using Microsoft.AspNetCore.Mvc.Razor;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


samples/Localization.StarterWeb/Startup.cs
```diff
@@ -15,6 +15,7 @@
 using Microsoft.AspNetCore.Localization;
 using System.Globalization;
 using Microsoft.Extensions.Options;
+using System.IO;
```
davidfowl: nit: Sort usings

samples/Localization.StarterWeb/Views/Shared/_Layout.cshtml
```diff
@@ -54,7 +54,7 @@
 
         <script src="//ajax.aspnetcdn.com/ajax/jquery/jquery-2.1.4.min.js"></script>
         <script src="//ajax.aspnetcdn.com/ajax/bootstrap/3.3.5/bootstrap.min.js"></script>
-        <script src="~/js/site.min.js" asp-append-version="true"></script>
+        <script src="~/js/site.js" asp-append-version="true"></script>
```
muratg: Why not minimized version?

ryanbrandenburg: The minimized version is empty and the unminified version is only like 3 lines long in the first place. @shanselman could you remove site.min.js while you're at it?

samples/Localization.StarterWeb/Startup.cs
```diff
@@ -1,19 +1,20 @@
 using System;
```
ryanbrandenburg: While we're messing around with the usings we may as well remove the couple unused ones in here.

ryanbrandenburg:
Other than some small nit-picks this looks like a :shipit:. Thanks for catching it!

Eilon:
@ryanbrandenburg can you make sure this PR is still valid and merge (or fix if needed)?

ryanbrandenburg:
Merged in 802ca6b75.

PR #158, opened 2016-08-04
Temporary fix for build break

kichalla:
@pranavkm as discussed offline this is a temporary fix until we fix the deployers in our testing to supply the working directory when running the app

@shanselman Your fix in the latest PR was correct but we have a bug in our end-to-end testing code where we do not handle the current working directory because of which tests are failing. I believe your original intention was to be able to do `dotnet run` on the app and be able to view the application. The fix in this PR makes that scenario work.  This is just a temp fix until we fix the deployer.


pranavkm:
:shipit: once the build passes.

PR #160, opened 2016-08-17
Uncoment InlineData and skip effected tests

### LINKED ISSUE #157: Fix and restore commented out tests
There are a couple of tests which have had their `InlineData`'s commented out. Restore those and resolve any failures. At worst if we can't resolve the underlying issue at this time we can skip, which has the benefit of being easier to tell that something isn't being tested.

ryanbrandenburg:
Fixes #157. Now we'll at least see these tests as skipped in test runs.

### DIFF
--- a/test/FunctionalTests/BuilderTests/BuilderFilteringWebTests.cs
+++ b/test/FunctionalTests/BuilderTests/BuilderFilteringWebTests.cs
@@ -20,15 +20,15 @@ namespace EntropyTests.BuilderTests
         [ConditionalTheory]
         [OSSkipCondition(OperatingSystems.Linux)]
         [OSSkipCondition(OperatingSystems.MacOSX)]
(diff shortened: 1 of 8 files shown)


muratg:
@ryanbrandenburg do we know if the reasons are still valid? i.e. are the tests failing because of the said condition in each test? I'm OK with disabling these, but I'd like to make sure that we're not just hiding unrelated issues.

ryanbrandenburg:
The reasons for all of these appear to still be valid, all the ones with x86 are blocked by [this](https://github.com/aspnet/Hosting/blob/f8189a2296e56b6b497afb98d49b004e283a93e5/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs#L40) and dotnet/corefx#9012 is still open and causing other tests to fail.

muratg:
Cool, thanks. :shipit:

PR #162, opened 2016-08-19
Do not test CLR on unix

JunTaoLuo:
cc @ryanbrandenburg. Currently the tests are broken on *nix systems. We don't want to test clr on *nix.


ryanbrandenburg:
Looks :shipit: to me.

PR #164, opened 2016-08-29
Added text/plain input formatter sample

kichalla:
@rynowak @javiercn

### DIFF
--- a/samples/Mvc.Formatters/Mvc.Formatters/Views/Home/Index.cshtml
+++ b/samples/Mvc.Formatters/Mvc.Formatters/Views/Home/Index.cshtml
@@ -1,0 +1,36 @@
+﻿@{
+    ViewData["Title"] = "Posting data in text/plain format";
+}
+<h2>Posting data in text/plain format</h2>
+<div>
+    <label>Sample text:</label>
+    <input type="text" id="sampleText" />
+    <button id="submitText" onclick="javascript: submitData()">Submit</button>
+    <div id="divStatus" style="margin-top:20px">
+        <label style="font-weight:bold">Echoed value: </label>
+        <label id="lblStatus"></label>
+    </div>
+</div>
+@section Scripts{
+    <script src="https://code.jquery.com/jquery-3.1.0.min.js"></script>
+    <script type="text/javascript">
+        $(function () {
+            $("#divStatus").hide();
+        });
+
+        function submitData() {
+            $.ajax({
+                type: "POST",
+                url: '/api/values',
+                data: $('#sampleText').val(),
+                contentType: 'text/plain',
+            }).done(function (data) {
+                $('#lblStatus').text(data);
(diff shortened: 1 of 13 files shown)


kichalla:
ping

kichalla:
ping

javiercn:
I'll look at it tomorrow

samples/Mvc.Formatters/Mvc.Formatters/Views/Home/Index.cshtml
```diff
@@ -0,0 +1,36 @@
...
+    <input type="text" id="sampleText" />
+    <button id="submitText" onclick="javascript: submitData()">Submit</button>
+    <div id="divStatus" style="margin-top:20px">
+        <label style="font-weight:bold">Echoed value: </label>
+        <label id="lblStatus"></label>
+    </div>
+</div>
+@section Scripts{
+    <script src="https://code.jquery.com/jquery-3.1.0.min.js"></script>
+    <script type="text/javascript">
```
javiercn: Can we do this with a regular form post instead of using JavaScript? It doesn't add value and just distracts customers from the important part of the sample.

samples/Mvc.Formatters/Mvc.Formatters/Views/Home/Index.cshtml
```diff
@@ -0,0 +1,36 @@
+﻿@{
+    ViewData["Title"] = "Posting data in text/plain format";
+}
+<h2>Posting data in text/plain format</h2>
+<div>
+    <label>Sample text:</label>
+    <input type="text" id="sampleText" />
+    <button id="submitText" onclick="javascript: submitData()">Submit</button>
```
javiercn: nit: onclick is from the nineties. It's considered a bad practice these days to do that. You can do 
```
$(function(){
  $('#submitText').on('click', submitData);
});
```

If you keep the JavaScript approach, at least change it to be from this century :smile:

javiercn:
:shipit: but either just use a regular form post or update the code to remove onclick

samples/Mvc.Formatters/Mvc.Formatters/Views/Home/Index.cshtml
kichalla: Yeah I actually did try that with `enctype=text/plain` in the form element and was expecting only the input field value to be sent over the wire but i see that it still goes as key-value pair. I didn't want this to be shown in the sample to avoid confusing with form url encoded format which typically is considered a key-value format type data.

javiercn: Nope, it will always send name/value pairs.

javiercn:
:shipit:

kichalla:
7a4ca4bfedebdffca133bc6ff10b4ccc6dc7a834

PR #168, opened 2016-09-27
Example of a custom ValidationAttribute

ryanbrandenburg:
Add an example of custom `ValidationAttribute` localization to` Mvc.LocalizationSample.Web`.

### DIFF
--- a/samples/Mvc.LocalizationSample.Web/CustomValidationAttributeAdapterProvider.cs
+++ b/samples/Mvc.LocalizationSample.Web/CustomValidationAttributeAdapterProvider.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.ComponentModel.DataAnnotations;
+using Microsoft.AspNetCore.Mvc.DataAnnotations;
+using Microsoft.AspNetCore.Mvc.DataAnnotations.Internal;
+using Microsoft.Extensions.Localization;
+
+namespace Mvc.LocalizationSample.Web
+{
+    public class CustomValidationAttributeAdapterProvider
+        : ValidationAttributeAdapterProvider, IValidationAttributeAdapterProvider
+    {
+        public CustomValidationAttributeAdapterProvider()
+        {
+        }
+
+        IAttributeAdapter IValidationAttributeAdapterProvider.GetAttributeAdapter(
+            ValidationAttribute attribute,
+            IStringLocalizer stringLocalizer)
+        {
+            var adapter = base.GetAttributeAdapter(attribute, stringLocalizer);
+
+            if (adapter == null)
+            {
+                var minLengthSix = attribute as MinLengthSixAttribute;
+                if (minLengthSix != null)
+                {
+                    adapter = new MinLengthSixAttributeAdapter(minLengthSix, stringLocalizer);
+                }
+            }
+
+            return adapter;
+        }
+    }
+}
--- a/samples/Mvc.LocalizationSample.Web/MinLengthSixAttribute.cs
+++ b/samples/Mvc.LocalizationSample.Web/MinLengthSixAttribute.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.ComponentModel.DataAnnotations;
+using System.Globalization;
+
+namespace Mvc.LocalizationSample.Web
+{
+    public class MinLengthSixAttribute : ValidationAttribute
+    {
+        public MinLengthSixAttribute()
+        {
+        }
+
+        public override string FormatErrorMessage(string name)
+        {
+            return string.Format(CultureInfo.CurrentCulture, ErrorMessageString, name);
+        }
+
+        public override bool IsValid(object value)
+        {
+            var stringValue = value as string;
+            if (stringValue == null)
+            {
+                return false;
+            }
+
+            return stringValue.Length >= 6;
+        }
+    }
+}
--- a/samples/Mvc.LocalizationSample.Web/MinLengthSixAttributeAdapter.cs
+++ b/samples/Mvc.LocalizationSample.Web/MinLengthSixAttributeAdapter.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Mvc.DataAnnotations.Internal;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+using Microsoft.Extensions.Localization;
+
+namespace Mvc.LocalizationSample.Web
+{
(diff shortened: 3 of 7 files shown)


samples/Mvc.LocalizationSample.Web/CustomValidationAttributeAdapterProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.AspNetCore.Mvc.DataAnnotations.Internal;
+using Microsoft.Extensions.Localization;
+
+namespace Mvc.LocalizationSample.Web
+{
+    public class CustomValidationAttributeAdapterProvider : IValidationAttributeAdapterProvider
+    {
+        public IAttributeAdapter GetAttributeAdapter(ValidationAttribute attribute, IStringLocalizer stringLocalizer)
+        {
+            var defaultProvider = new ValidationAttributeAdapterProvider();
```
dougbu: This approach couples the replacement provider with MVC `.Internal` implementation details. Suggest instead registering a `Func` which calls `serviceProvider.GetRequiredService<IValidationAttributeAdapterProvider>()` and passes that to a new instance of this provider.

samples/Mvc.LocalizationSample.Web/NotEqualToAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            return string.Format(CultureInfo.CurrentCulture, ErrorMessageString, name, Convert.ToString(Value));
+        }
+
+        public override bool IsValid(object value)
+        {
+            return !string.Equals(Convert.ToString(Value), Convert.ToString(value), StringComparison.CurrentCultureIgnoreCase);
+        }
+
+        public object Value { get; set; }
```
dougbu: - No reason for this to be settable.
- Move above the methods.

samples/Mvc.LocalizationSample.Web/NotEqualToAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        public override string FormatErrorMessage(string name)
+        {
+            return string.Format(CultureInfo.CurrentCulture, ErrorMessageString, name, Convert.ToString(Value));
+        }
+
+        public override bool IsValid(object value)
+        {
+            return !string.Equals(Convert.ToString(Value), Convert.ToString(value), StringComparison.CurrentCultureIgnoreCase);
```
dougbu: Could simplify this and `FormatErrorMessage` doing something like [`[Phone]`](https://github.com/dotnet/corefx/blob/master/src/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/PhoneAttribute.cs): If the value cannot be cast to `string`, it's not valid.

samples/Mvc.LocalizationSample.Web/NotEqualToAttributeAdapter.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Mvc.DataAnnotations.Internal;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+using Microsoft.Extensions.Localization;
+
+namespace Mvc.LocalizationSample.Web
+{
+    internal class NotEqualToAttributeAdapter : AttributeAdapterBase<NotEqualToAttribute>
```
dougbu: Why would an app ever have an `internal` class?

samples/Mvc.LocalizationSample.Web/NotEqualToAttributeAdapter.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        public override void AddValidation(ClientModelValidationContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            MergeAttribute(context.Attributes, "data-val", "true");
```
dougbu: What client-side validation will this actually enable? I suspect some JavaScript code is missing from this sample for the scenario to work end-to-end.

ryanbrandenburg: There is no JavaScript and therefore no client-side validation in the existing samples, this is just to make sure that anybody looking at this knows that that's the pattern. I've refactored it a bit to give a simpler idea of what you should do at the MergeAttribute part.

ryanbrandenburg:
:up::date:

samples/Mvc.LocalizationSample.Web/CustomValidationAttributeAdapterProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        IAttributeAdapter IValidationAttributeAdapterProvider.GetAttributeAdapter(
+            ValidationAttribute attribute,
+            IStringLocalizer stringLocalizer)
+        {
+            var adapter = GetAttributeAdapter(attribute, stringLocalizer);
+
+            if (adapter == null)
+            {
+                var type = attribute.GetType();
```
dougbu: Why is the `Type` needed? Suggest
``` c#
var minLengthSix = attribute as minLengthSix;
if (minLengthSix != null)
{
    adapter = new MinLengthSixAdapter(minLengthSix, stringLocalizer);
}
```

samples/Mvc.LocalizationSample.Web/CustomValidationAttributeAdapterProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        public CustomValidationAttributeAdapterProvider()
+        {
+        }
+
+        IAttributeAdapter IValidationAttributeAdapterProvider.GetAttributeAdapter(
+            ValidationAttribute attribute,
+            IStringLocalizer stringLocalizer)
+        {
+            var adapter = GetAttributeAdapter(attribute, stringLocalizer);
```
dougbu: Use `base.` to make it obvious why this doesn't recurse.

samples/Mvc.LocalizationSample.Web/CustomValidationAttributeAdapterProvider.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.ComponentModel.DataAnnotations;
+using Microsoft.AspNetCore.Mvc.DataAnnotations;
+using Microsoft.AspNetCore.Mvc.DataAnnotations.Internal;
```
dougbu: Please open an issue about moving `ValidationAttributeAdapterProvider` and `AttributeAdapterBase<TAttribute>` out of `.Internal`

samples/Mvc.LocalizationSample.Web/MinLengthSixAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        public override bool IsValid(object value)
+        {
+            var stringValue = value as string;
+            if (stringValue == null)
+            {
+                return false;
+            }
+            else
```
dougbu: nit: No need for `else`

ryanbrandenburg:
:up::date:

PR #169, opened 2016-09-27
Fix Class Library example

ryanbrandenburg:
This is a fix for the broken tests around localization, which were caused by a change it which GetResourcePrefix overloads we used.

@BrennanConroy you seem to be running point on this so let me know if someone else should review.


BrennanConroy:
:shipit:

PR #175, opened 2016-10-17
Add example of self-containted ValidationAttribute

### LINKED ISSUE #173: Cover simpler localization approach for custom validation attributes in MVC localization sample
The [Mvc Localization Sample](https://github.com/aspnet/Entropy/tree/dev/samples/Mvc.LocalizationSample.Web) demonstrates creating a custom validation attribute adapter i.e. an `IValidationAttributeAdapterProvider` and the `IAttributeAdapter` implementation it returns. As mentioned in aspnet/Mvc#5402, it's usually easier to override `IsValid(object, ValidationContext)` and to implement `IClientModelValidator` in the custom `ValidationAttribute` itself.

Extend the sample to demonstrate this simpler approach as well.

ryanbrandenburg:
Fixes #173 by providing an example of a ValidationAttribute which doesn't need an AttributeAdapter.


samples/Mvc.LocalizationSample.Web/Resources/Models.Product.fr.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="MaxLengthTwo" xml:space="preserve">
+    <value>{0} doivent avoir une longueur d'au plus deux.</value>
```
dougbu: This is a strange pattern to encourage. If user wants any language but French, they'll see "MaxLengthTwo" as the error message. Why aren't the localization keys useful to end users?

samples/Mvc.LocalizationSample.Web/SelfContainedValidationAttribute.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.AspNetCore.Mvc.ModelBinding.Validation;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Localization;
+using Microsoft.Extensions.Options;
+
+namespace Mvc.LocalizationSample.Web
+{
+    public abstract class SelfContainedValidationAttribute : ValidationAttribute, IClientModelValidator
+    {
+        public abstract void AddValidation(ClientModelValidationContext context);
```
dougbu: Why?

samples/Mvc.LocalizationSample.Web/SelfContainedValidationAttribute.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        private IStringLocalizer GetStringLocalizer(
```
dougbu: Move `private` methods after `protected` ones.

samples/Mvc.LocalizationSample.Web/SelfContainedValidationAttribute.cs
```diff
@@ -0,0 +1,79 @@
...
+                string.IsNullOrEmpty(ErrorMessageResourceName) &&
+                ErrorMessageResourceType == null)
+            {
+                return stringLocalizer[ErrorMessage, displayName];
+            }
+
+            return FormatErrorMessage(displayName);
+        }
+
+        protected static bool MergeAttribute(IDictionary<string, string> attributes, string key, string value)
```
dougbu: Can kind-off understand splitting `SelfContainedValidationAttribute` out in general. But why isn't this just a `private` method in `MaxLengthTwoAttribute`? It's unrelated to other methods in this class.

samples/Mvc.LocalizationSample.Web/MaxLengthTwoAttribute.cs
```diff
@@ -0,0 +1,56 @@
...
+        public MaxLengthTwoAttribute()
+        {
+            ErrorMessage = _errorMessage;
+        }
+
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            var result = base.IsValid(value, validationContext);
+
+            if (result.ErrorMessage != null)
```
ryanbrandenburg: I had explicitly done `!= null` because I thought it was valid for someone to set the ErrorMessage to the empty string in their attribute, but I can't find anything to support that, so sure.

samples/Mvc.LocalizationSample.Web/Resources/Models.Product.fr.resx
ryanbrandenburg: This is definitly not the patern we want to encourage, but it's also the format the other model resources are using. I filed #176 to resolve for all instances where it's happened and keep this PR relativly on subject.

samples/Mvc.LocalizationSample.Web/SelfContainedValidationAttribute.cs
ryanbrandenburg: Because otherwise `SelfContainedValidationAttribute` wouldn't completly implement `IClientModelValidator`. Do you have another method of doing this that you recomend?

ryanbrandenburg:
:up::date:

samples/Mvc.LocalizationSample.Web/Resources/Models.Product.fr.resx
dougbu: :ok:

samples/Mvc.LocalizationSample.Web/SelfContainedValidationAttribute.cs
dougbu: Nah, that's fine. Was a question.

samples/Mvc.LocalizationSample.Web/SelfContainedValidationAttribute.cs
```diff
@@ -0,0 +1,70 @@
...
+            MvcDataAnnotationsLocalizationOptions options,
+            IStringLocalizerFactory factory,
+            Type modelType)
+        {
+            var provider = options.DataAnnotationLocalizerProvider;
+            if (factory != null && provider != null)
+            {
+                return provider(modelType, factory);
+            }
+            else
```
dougbu: Nit: No need for `else` after a `return`.

Issue #209, opened 2017-03-21
Container tests fail with EADDRINUSE


jbagga:
ContainerFallbackWebTests.KestrelX64CoreCLR failed on the Ubuntu test run
http://aspnetci/viewLog.html?buildId=190297&tab=buildResultsDiv&buildTypeId=XPlat_Linux_UbuntuUniverseCoreCLR

jbagga:
cc @halter73

halter73:
I have no idea what's going on with this one. The client simply failed to establish a TCP connection. Maybe Kestrel just took too long to start this one time? That's why there's a retry.

Even though TC marked it as flaky, if you look at the test history it looks like this is the only time this test has ever failed: http://aspnetci/project.html?tab=testDetails&projectId=XPlat&testNameId=8180339492066270673&page=1

jbagga:
[ConfigWalkingValuesWebTests.KestrelX64CoreCLR](https://github.com/aspnet/Entropy/blob/dev/test/FunctionalTests/ConfigTests/ConfigWalkingValuesWebTests.cs)
Failure details: http://aspnetci/viewLog.html?buildId=191998&tab=buildResultsDiv&buildTypeId=XPlat_Linux_UbuntuUniverseCoreCLR.

@halter73 This one has the same error. Fails to connect and retry is canceled

Eilon:
@halter73 we want to abolish all flaky tests. We have too many flaky tests across the board, so it's a death-by-a-thousand-paper-cuts situation.

muratg:
@natemcmaster could you take a look? (@halter73 is out this week.)

natemcmaster:
ContainerFallbackWebTests failed with

> Unhandled Exception: System.IO.IOException: Failed to bind to address http://localhost:6000 on the IPv4 loopback interface: port already in use. ---> System.AggregateException: One or more errors occurred. (Error -98 EADDRINUSE address already in use) ---> Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvException: Error -98 EADDRINUSE address already in use

ConfigWalkingValuesWebTests failed with the same error, but ports 5900. I couldn't find any other tests that explicitly use ports 5900 or 6000. Could this be an issue in Kestrel?

Eilon:
Didn't we have some automatic port selector thingy that we used somewhere for Kestrel tests? Or some other tests?

@mikeharder @Tratcher maybe you guys remember?

natemcmaster:
[...]

PR #216, opened 2017-03-28
Use dynamic ports and replace console logging with xunit logger in E2E tests

natemcmaster:
Addresses #209.

Use dynamic ports instead of fixed ports.
Switch from using console logging to using the xunit logger. Remove duplicated code.


test/FunctionalTests/E2ETestBase.cs
```diff
@@ -15,83 +20,80 @@ public abstract class E2ETestBase
         private string _siteName;
         private ITestOutputHelper _output;
 
-        protected E2ETestBase(ITestOutputHelper output, string siteName, int basePort)
+        protected E2ETestBase(ITestOutputHelper output, string siteName)
         {
-            _output = output;
+            _output = output ?? throw new ArgumentNullException(nameof(output));
```
Eilon: No throw expressions please.

test/FunctionalTests/E2ETestBase.cs
```diff
@@ -141,5 +142,19 @@ protected virtual void AssertResponse(HttpResponseMessage response, string respo
         {
             throw new NotImplementedException("Must be overriden in derived types.");
         }
+
+        private static int GetNextPort()
```
Eilon: I assume this code is copied from some other place? Is it worth putting it in one of our test helper classes?

test/FunctionalTests/XunitLogger.cs
```diff
@@ -1,61 +0,0 @@
...
-// See License.txt in the project root for license information
-
-using System;
-using System.Linq;
-using Microsoft.Extensions.Logging;
-using Xunit.Abstractions;
-
-namespace EntropyTests
-{
-    public class XunitLogger : ILogger, IDisposable
```
Eilon: Nice.

test/FunctionalTests/E2ETestBase.cs
natemcmaster: Because of preference or technical reasons?

test/FunctionalTests/E2ETestBase.cs
natemcmaster: Yes copied. Probably not worth maintaining as an API. It's simple enough and it's just test code.

test/FunctionalTests/E2ETestBase.cs
Eilon: Engr team decided they are evil.

natemcmaster:
:up: :date: no-mo-throw expressions

test/FunctionalTests/E2ETestBase.cs
anurse: Similar code actually already existed in the IntegrationTests package from Hosting. I've converted the Hosting version to be a copy of the Kestrel one and some of this code will be able to be removed once aspnet/Hosting#996 lands :).

PR #217, opened 2017-03-30
react to aspnet/Hosting#996

anurse:
The saga continues.

@natemcmaster , I removed your helper since we now have this functionality built in to the IntegrationTesting library.

Related to aspnet/Hosting#996

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -1,6 +1,6 @@
-<Project>
+﻿<Project>
   <PropertyGroup>
-    <AspNetCoreLabsVersion>0.3.0-*</AspNetCoreLabsVersion>
+    <AspNetIntegrationTestingVersion>0.4.0-*</AspNetIntegrationTestingVersion>
     <AspNetCoreVersion>1.2.0-*</AspNetCoreVersion>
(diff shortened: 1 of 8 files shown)


build/dependencies.props
```diff
@@ -1,6 +1,6 @@
-<Project>
+﻿<Project>
   <PropertyGroup>
-    <AspNetCoreLabsVersion>0.3.0-*</AspNetCoreLabsVersion>
+    <AspNetIntegrationTestingVersion>0.4.0-*</AspNetIntegrationTestingVersion>
```
JunTaoLuo: I realize this may be a little late but why did we bump the version for the IntegrationTesting package?

anurse: To avoid breaking the world when we introduced breaking changes: https://github.com/aspnet/Hosting/pull/996#issuecomment-290195195

JunTaoLuo: Huh, I guess this is fine since this is not a package release but I still find it strange that we start bumping version numbers twice for the 2.0 milestone.

anurse: This is a manually-versioned package for exactly this reason. It's never shipped.

PR #234, opened 2017-05-04
netcoreapp2.0

mikeharder:


test/Entropy.FunctionalTests/MvcTests/LocalizationSampleTest.cs
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-#if NET46
```
JunTaoLuo: The ifdef here suggests the tests were only meant for net46 so why are we keeping it? If it runs it's fine but then I'd be curious why it wasn't run before.

pranavkm: This was ifdefed because localization was broken on netcoreapp at some point but never reverted.

PR #248, opened 2017-06-07
Re-enable tests on net461

mikeharder:


mikeharder:
> These tests were already trying to simulate `<PreserveCompilationContext>` here: https://github.com/aspnet/Entropy/blob/3acb444f0b49067b5fa45b5ca31538b16c099a1c/test/Entropy.FunctionalTests/Entropy.FunctionalTests.csproj#L66-L76.  Instead of that copying code, the tests can just set `<PreserveCompilationContext>` and these errors should go away.

@eerhardt: Were you suggesting we should be able to remove the `CopyDepsFile` target after setting `<PreserveCompilationContext>true</PreserveCompilationContext>`?  I tried this and it caused tests to fail with errors like:

```
D:\Git\Entropy\samples\Mvc.TagHelperSample.Web\Views\Home\Index.cshtml(50,21): error CS0518: Predefined type 'System.String' is not defined or imported
```

All tests pass on my dev machine with both `<PreserveCompilationContext>true</PreserveCompilationContext>` and the `CopyDepsFile` target.  I will plan to merge this change for Preview2, and file an issue for post-Preview2 to investigate whether the `CopyDepsFile` target can be removed.

NTaylorMullen:
> @eerhardt: Were you suggesting we should be able to remove the CopyDepsFile target after setting <PreserveCompilationContext>true</PreserveCompilationContext>?

No, you need both.

eerhardt:
When I first glanced, I thought it was copying just a single .deps.json file.  But now I see this one test project references a bunch of web projects, and is copying multiple .deps.json files into the test folder.  PreserveCompilationContext won't do that.  So I was wrong above, you will still need to do the `CopyDepsFile` step.

mikeharder:
OK, current changes pass on my dev machine, will merge once AppVeyor is green.  Travis is failing in dev branch as well.

mikeharder:
Addresses https://github.com/aspnet/Coherence-Signed/issues/540

Issue #260, opened 2017-08-21
Update RenderViewToString to use `GetView` \ `FindView` fallback


pranavkm:
We should modify the sample to use the same pattern for view discovery as `ViewExecutor`: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/ViewResultExecutor.cs#L79-L84.

PR #261, opened 2017-08-23
Use GetView in addition to FindView for view discovery

pranavkm:
Fixes https://github.com/aspnet/Entropy/issues/260


samples/Mvc.RenderViewToString/RazorViewToStringRenderer.cs
```diff
@@ -34,15 +36,7 @@ public class RazorViewToStringRenderer
         public async Task<string> RenderViewToStringAsync<TModel>(string name, TModel model)
```
ajaybhargavb: Could you rename the param to `viewPath`? Having `name` doesn't make sense.

pranavkm: On second thought, the current name matches what we have in `ViewResult`: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/ViewResult.cs#L30. I'll change it to `viewName` and call it a day (particularly given that it's not necessarily a path all the time).

ajaybhargavb: Sure. My concern was mostly about having to specify the `.cshtml` part of the path.

pranavkm: We don't always require the extension `RenderViewToStringAsync("View", null)` would work just as fine. However, there are oddities with the paths we generate that prevent it from working with precompiled views. If / when we choose to fix it, we should be able to use either interchangeably.

PR #262, opened 2017-08-29
Add Controller for multiple areas sample

jbagga:
Sample code for aspnet/Mvc#6637

cc @rynowak @Eilon


samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasControllerConvention.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace ControllerForMultipleAreasSample
+{
+    public class MultipleAreasControllerConvention : IApplicationModelConvention
+    {
+        public MultipleAreasControllerConvention()
+        {
+            ControllerModels = new List<ControllerModel>();
+        }
+
+        public IList<ControllerModel> ControllerModels { get; set; }
```
rynowak: This should be a local variable, there's no need for it to exist outside the scope of `Apply`

samples/Mvc/Mvc.ControllerForMultipleAreasSample/Startup.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
```
rynowak: You seem to have two copies of the files:
```
samples/Mvc.ControllerForMultipleAreasSample/Startup.cs
samples/Mvc/Mvc.ControllerForMultipleAreasSample/Startup.cs
```

samples/Mvc/Mvc.ControllerForMultipleAreasSample/Mvc.ControllerForMultipleAreasSample.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Folder Include="wwwroot\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
+  </ItemGroup>
```
rynowak: This should look like our other projects in this repo, and use the latest bits

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasControllerConvention.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public void Apply(ApplicationModel application)
+        {
+            foreach (var controller in application.Controllers)
+            {
+                var areas = controller.ControllerType.GetCustomAttributes<MultipleAreasAttribute>().FirstOrDefault();
+                var areaNames = areas?.AreaNames;
```
rynowak: it would be clearer if you just short circuited this entirely

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasControllerConvention.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Microsoft.AspNetCore.Mvc.ApplicationModels;
+
+namespace ControllerForMultipleAreasSample
+{
+    public class MultipleAreasControllerConvention : IApplicationModelConvention
```
rynowak: You can also use `IControllerModelConvention` instead and then you don't need the loop.

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace ControllerForMultipleAreasSample
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class MultipleAreasAttribute : Attribute
+    {
+        public MultipleAreasAttribute(string[] areaNames)
```
rynowak: a handy pattern for these things is to do something like `MulipleAreasAttribute(string a1, string a2, params string[] a3...)` that makes it clear you're supposed to use at least 2 values.

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasControllerConvention.cs
```diff
@@ -0,0 +1,46 @@
...
+        public IList<ControllerModel> ControllerModels { get; set; }
+
+        public void Apply(ApplicationModel application)
+        {
+            foreach (var controller in application.Controllers)
+            {
+                var areas = controller.ControllerType.GetCustomAttributes<MultipleAreasAttribute>().FirstOrDefault();
+                var areaNames = areas?.AreaNames;
+
+                for (var i = 0; i < areaNames?.Length; i++)
```
rynowak: a more common pattern for this is to just start the loop at 1. So move L28 outside of the loop, and then simplify the body to just L33-35

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace ControllerForMultipleAreasSample
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
```
Eilon: I think this attribute can target only classes, right? Not methods?

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+    public class MultipleAreasAttribute : Attribute
+    {
+        public MultipleAreasAttribute(string[] areaNames)
+        {
+            AreaNames = areaNames;
+        }
+
+        public string[] AreaNames { get; private set; }
```
Eilon: Probably don't need `private set` since this is set only in the ctor.

samples/Mvc.ControllerForMultipleAreasSample/Areas/Products/Controllers/HomeController.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.AspNetCore.Mvc;
```
Eilon: Add license headers to all the files too.

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasControllerConvention.cs
jbagga: How will the new copied controllers with different area value in its RouteValues be added to the application?

jbagga:
@Eilon @rynowak Updated

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasControllerConvention.cs
rynowak: Not explained here but we ran into a problem with this approach. You can't modify `application.Controllers` while inside an `IControllerModelConvention`.

samples/Mvc.ControllerForMultipleAreasSample/Areas/Products/Controllers/HomeController.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Mvc;
+
+namespace ControllerForMultipleAreasSample.Areas.Products.Controllers
```
rynowak: It might be clearer to put this somewhere else. This could just go in the root (file and namespace wise)

samples/Mvc.ControllerForMultipleAreasSample/Areas/Services/Views/Home/Index.cshtml
```diff
@@ -0,0 +1 @@
+﻿Services View
```
rynowak: Need a view for `Manage`?

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasControllerConvention.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Collections.Generic;
```
rynowak: license

samples/Mvc.ControllerForMultipleAreasSample/Areas/Services/Views/Home/Index.cshtml
jbagga: https://github.com/aspnet/Entropy/pull/262/files#diff-e9ab401e3324db45ac2bc4aa3f860381

jbagga:
@Eilon Updated

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace ControllerForMultipleAreasSample
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
+    public class MultipleAreasAttribute : Attribute
+    {
+        public MultipleAreasAttribute(string area1, string area2, params string[] areaNames)
+        {
+            AreaNames = new string[] { area1, area2 };
+            AreaNames = AreaNames.Concat(areaNames).ToArray();
```
Eilon: Can probably do both the above lines in 1 line of code.

samples/Mvc.ControllerForMultipleAreasSample/MultipleAreasControllerConvention.cs
```diff
@@ -0,0 +1,34 @@
...
+                {
+                    var controllerCopy = new ControllerModel(controller);
+                    controllerCopy.RouteValues["area"] = areaNames[i];
+                    controllerModels.Add(controllerCopy);
+                }
+            }
+
+            foreach (var model in controllerModels)
+            {
+                application.Controllers.Add(model);
```
Eilon: Any reason not to do this within the inner nested loop above? That is, instead of making a list, then iterating over the list, why not just add each new model immediately?

jbagga: Because that modifies the collection (application.Controllers), which halts the enumeration (in the above for loop).

Eilon: Oh, obviously, thanks! 😄

jbagga:
@Eilon Added a README

PR #263, opened 2017-08-29
Upgrade xunit

ryanbrandenburg:
This was causing a failure on the CI.


NTaylorMullen:
- Why'd upping the xunit version fix?
- What was the failure?

ryanbrandenburg:
The answer to both is [here](http://aspnetci/viewLog.html?buildId=316773&buildTypeId=Lite_Entropy&tab=buildLog&_focus=949). Pretty straightforward instance of package mismatch.

PR #264, opened 2017-09-11
[Docs] Add Cookie auth sample

HaoK:
PR for new doc samples, won't check these into Entropy


samples/Authentication.Cookies/Startup.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            services.AddAuthentication("Cookies") // Sets the default scheme to cookies
+                .AddCookie("Cookies", options =>
+                {
+                    options.AccessDeniedPath = "/account/denied";
+                    options.LoginPath = "/account/login";
+                });
+
+            // claims transformation is run after every Authenticate call
+            services.AddTransient<IClaimsTransformation, ClaimsTransformer>();
```
davidfowl: Should these be in the same sample?

HaoK: I can create a stand alone ClaimsTransformation sample instead, I'll split these out

HaoK:
Replaced by https://github.com/aspnet/AuthSamples/pull/1

PR #267, opened 2017-10-18
Fix Xunit version mismatch

ryanbrandenburg:
This Xunit didn't match the one used in one of the other projects, so it was failing on the CI.


ryanbrandenburg:
Tests are failing, but at a later point.

PR #271, opened 2017-11-16
Use Directory.Build.props

ryanbrandenburg:
This allows Entropy to build locally. There's a test failure, but I'm reasonably sure it was pre-existing.


version.props
```diff
@@ -0,0 +1,11 @@
+<!-- This file may be overwritten by automation. -->
+<Project>
+  <PropertyGroup>
+    <VersionPrefix>2.1.0</VersionPrefix>
```
pranavkm: There isn't any reason to version artifacts from this repo. We shouldn't need this.

ryanbrandenburg:
Weird, it passed locally. Fixing.

version.props
ryanbrandenburg: If we remove all the version prefix stuff KoreBuild complains that the version `1.0.0` doesn't match `2.1.0-preview1-t000`. It doesn't really matter if the versions are "right" on this repo, but it doesn't hurt anything that they're there, so I don't see a need to circumvent KoreBuild.

PR #272, opened 2017-11-16
Fix JSON patch test

jbagga:
@ryanbrandenburg Can you try building it locally? I could not run Entropy tests locally.

Or I can wait for https://github.com/aspnet/Entropy/pull/271 to be checked in and then merge this in once I know it passes.


ryanbrandenburg:
Yeah, you might just wait on #271, that way we can be sure it fixes it.

jbagga:
@ryanbrandenburg Can you restart the Mac tests?

ryanbrandenburg:
I restarted it, but if the only failure you get is the stupid CuRL OSX only thing then feel free to merge, Nate and I are considering not even running OSX in Travis since it's so unreliable.

jbagga:
@ryanbrandenburg Same error again. Merging now

PR #275, opened 2018-02-15
Tweak the Mvc.CustomRazorPageHandlers samp

### LINKED ISSUE #270: Mvc.CustomRazorPageHandlers sample update?
RE: Mvc.CustomRazorPageHandlers sample

https://github.com/aspnet/Entropy/tree/dev/samples/Mvc.CustomRazorPageHandlers

I couldn't get the current implementation to work for the sample I built to accompany the [Razor Pages route and app convention features (aspnet/docs 4623)](https://github.com/aspnet/Docs/pull/4623) topic. I modified (some would say since I did it ... *"hacked"* :smile:) it to work with the code below.

~I can't PR this into the repo. I'm on a strict contract that prevents me from submitting PR's outside of the docs repo right now. When the contract ends early next year if the sample still needs an update, then I'll swing over and take care of this if u like.~

cc/ @pranavkm 

```csharp
using System;
using System.Reflection;
using Microsoft.AspNetCore.Mvc.ApplicationModels;
using Microsoft.AspNetCore.Mvc.RazorPages.Internal;

namespace ModelProvidersSample
{
    public class CustomPageApplicationModelProvider : DefaultPageApplicationModelProvider
    {
        protected override PageHandlerModel CreateHandlerModel(MethodInfo method)
        {
            if (method == null)
            {
                throw new ArgumentNullException(nameof(method));
            }

            if (!IsHandler(method))
            {
                return null;
            }

            if (!TryParseHandlerMethod(
                method.Name, out var httpMethod, out var handlerName))
            {
                return null;
            }

            var handlerModel = new PageHandlerModel(
                method,
                method.GetCustomAttributes(inherit: true))
            {
                Name = method.Name,
                HandlerName = handlerName,
                HttpMethod = httpMethod,
            };

            var methodParameters = handlerModel.MethodInfo.GetParameters();

            for (var i = 0; i < methodParameters.Length; i++)
            {
                var parameter = methodParameters[i];
                var parameterModel = CreateParameterModel(parameter);
                parameterModel.Handler = handlerModel;

                handlerModel.Parameters.Add(parameterModel);
            }

            return handlerModel;
        }

        private static bool TryParseHandlerMethod(
            string methodName, out string httpMethod, out string handler)
        {
            httpMethod = null;
            handler = null;

            // Parse the method name according to our conventions to 
            // determine the required HTTP verb and optional 
            // handler name.
            //
            // Valid names look like:
            //  - Get
            //  - Post
            //  - PostAsync
            //  - GetMessage
            //  - PostMessage
            //  - DeleteMessage
            //  - DeleteMessageAsync

            var length = methodName.Length;
            if (methodName.EndsWith("Async", StringComparison.Ordinal))
            {
                length -= "Async".Length;
            }

            if (length == 0)
            {
                // The method is named "Async". Exit processing.
                return false;
            }

            // The HTTP verb is at the start of the method name. Use 
            // casing to determine where it ends.
            var handlerNameStart = 1;
            for (; handlerNameStart < length; handlerNameStart++)
            {
                if (char.IsUpper(methodName[handlerNameStart]))
                {
                    break;
                }
            }

            httpMethod = methodName.Substring(0, handlerNameStart);

            // The handler name follows the HTTP verb and is optional. 
            // It includes everything up to the end excluding the 
            // "Async" suffix, if present.
            handler = handlerNameStart == length ? null : methodName.Substring(0, length);

            if (string.Equals(httpMethod, "GET", StringComparison.OrdinalIgnoreCase) || 
                string.Equals(httpMethod, "POST", StringComparison.OrdinalIgnoreCase))
            {
                // Do nothing. The httpMethod is correct for GET and POST.
                return true;
            }
            if (string.Equals(httpMethod, "DELETE", StringComparison.OrdinalIgnoreCase) || 
                string.Equals(httpMethod, "PUT", StringComparison.OrdinalIgnoreCase) || 
                string.Equals(httpMethod, "PATCH", StringComparison.OrdinalIgnoreCase))
            {
                // Convert HTTP verbs for DELETE, PUT, and PATCH to POST
                // For example: DeleteMessage, PutMessage, PatchMessage -> POST
                httpMethod = "POST";
                return true;
            }
            else
            {
                return false;
            }
        }
    }
}
```

guardrex:
Fixes #270 

Not broken after all. I just suggest a couple of minor tweaks to the page functions.

Since the problem I experienced might have been with the implementation of this `TryParseHandlerMethod` in the doc sample, I'll go back and double-check the sample and make sure that it works properly.


pranavkm:
Thanks for the PR!

ryanbrandenburg:
@pranavkm this PR causes CustomRazorPageHandlerTest.SimpleGet_Works to [always fail](https://ci.appveyor.com/project/aspnetci/entropy/build/1.0.767#L1551). Can we get a reaction to fix it?

guardrex:
https://github.com/guardrex/Entropy/blob/dev/test/Entropy.FunctionalTests/MvcTests/CustomRazorPageHandlerTest.cs#L28

PR #277, opened 2018-05-24
Update CustomRazorPageHandlerTest.cs

pranavkm:


pranavkm:
@ryanbrandenburg so that's not the only reason tests are failing. See Travis failure

PR #279, opened 2018-06-07
Add a sample showing partial page updates in Razor Pages

pranavkm:


pranavkm:
cc @DamianEdwards

samples/Mvc.RazorPagePartial/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,26 @@
...
+        public string Message => DateTime.Now.ToString();
+
+        public IActionResult OnGet() => Page();
+
+        public IActionResult OnGetPartial()
+        {
+            return new PartialViewResult
+            {
+                ViewName = "_IndexPartial",
+                ViewData = new ViewDataDictionary<IndexModel>(this.ViewData, this),
```
DamianEdwards: So this is the magic incantation. Would be nice to have a built-in helper for this of course as this isn't particularly discoverable (passing in two `this` members, plus the current type as the generic arg, eeep).

samples/Mvc.RazorPagePartial/Pages/Index.cshtml
```diff
@@ -0,0 +1,13 @@
+@page "{handler?}"
+@model IndexModel
+
+<div id="ajax-partial">
```
DamianEdwards: Is this effectively identical to how it works in MVC of yore?

samples/Mvc.RazorPagePartial/Pages/Index.cshtml.cs
pranavkm: Yup. Basically need https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Controller.cs#L158-L209 on PageModel

samples/Mvc.RazorPagePartial/Pages/Index.cshtml
pranavkm: As far as I can tell, yes. There's obviously niceties around generating the link and form `Ajax.BeginForm` \ `Ajax.ActionLink`, but the remainder of it looks to be pretty consistent with Mvc 5.

